PR #62, opened 2026-09-09
Statement transform for migrated React, and the 61 backend defects behind it

paraphilic-ecchymosis:
Work from the Lares-as-Sigil capability probe. The point is not a Sigil app — it is to find the gaps in Sigil and Qliphoth by building a real one, and the gaps are the deliverable. They are recorded as S63–S124 in `docs/specs/LARES-SIGIL-REBUILD-SPEC.md` in the Lares repository (spec 1.30.0).

**Base:** retargeted from `main` to `develop`, per #69 and the note in #74. See *Conflicts* at the bottom — this needs #74 to land first.

## What changed

**A statement walk over the JavaScript AST** (`migrate::react::stmt_transform`). Handler bodies had been reduced to a flat list of state assignments, so a handler that wrote one field on success and another on failure emitted both, unconditionally, in source order. Anything the walk cannot translate becomes a `// React:` comment carrying the original source, so a missing statement is visible rather than silently dropped.

Block placeholders 441 → 0, flattened handler notes 33 → 0.

**Then the defects behind it.** Translating handler bodies made 206 helper bodies reachable that had never run, and the whole client linkable for the first time. Each new thing that ran found the next one.

### The value model

- **`.await` parsed as a field named `await`.** The parser accepted `.⌛` and not `.await`, so `p.await` compiled to a load at offset 0 of a promise id. **Every `.await` in a migrated program is that spelling**, so not one of them awaited anything, in code that compiled and reported success.
- **An Option had two representations that disagreed.** `Some(x)` compiles to `x` and wraps nothing; `None` allocated a fresh 16-byte Option at every mention with equality on the pointer. So `None == None` was false, `∅ == None` was false, and `⌥ Some(5) { Some(v) => v }` gave 0 because the pattern dereferenced the payload. Nothing is a sentinel now, shared with the host and pinned by a test — not 0, which is also `false` and every host function's "absent".
- **Two bump allocators over one linear memory.** The module bumped `__heap_ptr` inline while the host kept its own at a constant; the client's literals reach past it. Shared now.
- **Closure arity is part of the WASM type**, so a one-parameter closure and a two-argument call site trapped in `call_indirect`. One signature for every closure.
- **JavaScript's `/` never fails and Sigil's is an error on zero.** `averageTrendTotal` divides by `trend.length` with no guard, exactly as its React does.
- **`Math.floor` could not be called at all** — `math.*` is F64-typed and every migrated number is an i64.
- **`Date.now()` was `performance.now()`**: milliseconds since page load, as an F64, so every age and staleness check measured epoch timestamps against a three-digit number.

### What reached the document

Found by rendering the client in a browser with the real stylesheet and looking at it — none of these is visible to a test that asks whether code runs:

- JSX attribute names are not DOM attribute names (`strokeWidth`, `htmlFor`, `tabIndex`), so a browser ignored them silently.
- `ref` and `key` reached the DOM as `ref="0"`.
- `className={chip.cls}` rendered the literal text `chip.cls`.
- A sibling component in a child position became text, so a VNode rendered as the decimal of its own address.
- Booleans rendered as `1`/`0`, including in `aria-expanded`.
- `None` rendered as a heap pointer where React renders nothing.
- **Every anonymous object shared one field layout** — the parser names them all `__anonymous__`, so the first literal defined the shape and every later one wrote its fields at offset 0. The client is built out of props bags.

### The migration's own reach

- **A file-private helper had no way to travel.** Helpers arrive through imports and the resolver followed only exported ones — while the constants resolver beside it had a second table for exactly this, with a comment explaining why. `isMrList` is exported and arrived; the `isMr` it calls and the `isObj` behind that did not. **`isObj` alone blocked 15 helpers.**
- The walk queued `h.calls`, which is empty on every extracted helper — and would have missed `v.every(isMr)` anyway, a bare reference rather than a call.
- The same shape for constants only a helper's body names.
- `let x; if (a) x = …; else x = …;` dropped every branch — the whole of what `RefreshIndicator` renders.
- Default parameter values were dropped, so `formatRelativeTime(ms, now = Date.now())` compared every timestamp against 0.

**157 untranslated helper bodies → 49.**

### Diagnostics

`SIGIL_MIGRATE_SKIPS` says why each helper was skipped. It was written before any fix here, and it is why: the 157 were not waiting on a statement transform — that existed and worked — they were translated and thrown away by a downstream gate, and 165 of 166 failed the same one. "expression is not constant" now names the expression, which is how a module-scope float with no const arm was found in a minute instead of by bisecting 3000 lines.

## Validation

- **885 unit tests** pass.
- **Sigil suite 794/798** — the 4 are the Kafka/AMQP environmental ones.
- **Website: 8/8 pages** compile from a clean build.
- **Lares client: 93/93 components** compile, the whole project links, and all 93 render in a browser.
- **Differential sweep 105/107.** The two failures are recorded findings, documented as such in `tools/differential/README.md`: float division, and `(None)·to_string()` being `"None"` in the interpreter and `""` in WASM.
- **Import contract 220/220** host functions checked against the types the module declares.

## Known, and not fixed here

**`==` is strictly typed in the interpreter and untyped in WASM** (S124). `∅ == None`, `false == None` and `0 == None` are runtime type errors under one backend and ordinary comparisons under the other; `∅` and `None` are not even the same type in the interpreter. A migrated component is made of `x == None` where `x` is `Any`, so the client runs — and has only ever run — in WASM. The Option change here lands in the WASM backend only; the interpreter, Cranelift and LLVM backends are untouched.

`Ok(x)` and `Err(x)` are both transparent, so a Result cannot tell its two sides apart. 49 helper bodies remain untranslated: browser globals with no Sigil equivalent (`localStorage`, `AbortController`, `URLSearchParams`), closure-captured factory state, destructuring.

## Conflicts

This branch is based on `main` and `develop` is 8 commits ahead of it, with 13 files conflicting — `interpreter.rs`, `parser.rs`, `stdlib.rs`, `llvm_codegen.rs` among them. **Those are the same files #74 is reconciling.** Resolving them here now would be throwaway work against a `develop` that #74 replaces, and risks contradicting resolutions #74 argues for in detail. The sequencing is: **#74 lands, then `develop` merges into this branch.**

https://claude.ai/code/session_01MgQA8aadFD4ztC8axmS5WN


paraphilic-ecchymosis:
**`Build and Test (ubuntu-latest, stable)` is red on `133370b`, and it is not this PR's.**

The job died 20 seconds in, during dependency install, before anything was built or run:

```
E: Failed to fetch https://dl.google.com/linux/chrome-stable/deb/dists/stable/main/binary-amd64/Packages.gz
   Hash Sum mismatch
E: Some index files failed to download.
##[error]Process completed with exit code 100.
```

Google's CDN was serving a Chrome package index whose hash did not match its own Release file (`Release file created at: Wed, 09 Sep 2026 17:16:59`, index last modified `09:41:12`). Neither this repository nor this PR uses that apt source; it is on the runner image.

Evidence it is not the diff:

- The identical job passed at 17:20 on `f4f10e3`, the previous commit on this branch.
- `Build and Test (macos-latest, stable)` is building normally on this same commit — macOS does not use apt.
- `Build MCP Server` passed on this same commit.
- The same failure hit Daemoniorum-LLC/qliphoth#10 at 17:53, in its `playwright install --with-deps` step.

Locally on this exact commit: 885 unit tests pass, the Sigil suite is 794/798 (the 4 Kafka/AMQP environmental ones), and all 8 website pages build clean.

**I cannot re-run it** — `rerun-failed-jobs` returns 403 for this integration. I have not patched the workflow to drop the Chrome apt source, because that is a CI-configuration decision for this repository rather than something this PR should carry, and the upstream index resolves itself.

Anyone with write access can re-run the failed job. Otherwise the next push to this branch will re-trigger it; I am keeping the PR watched until it is green.

---

PR #66, opened 2026-09-09
fix(parser): restore middledot module paths in type position (regression from #61)

paraphilic-ecchymosis:
## What

`std·fmt·Formatter<'_>` stopped parsing on `main`. **This is a regression from #61**, bisected to that commit.

`#61` tightened `parse_type_path`:

```rust
// a0180d3 (#59) and earlier — the first clause consumes `·` unconditionally,
// which makes the second clause dead code
let is_path_sep = self.consume_if(&Token::MiddleDot)
    || (first_segment_is_type && self.consume_if(&Token::MiddleDot));

// 8f92d5c (#61) — now gated on the first segment being uppercase
let is_path_sep = first_segment_is_type && self.consume_if(&Token::MiddleDot);
```

The tightening was intentional and its comment says why: `·` after a lowercase segment "must be left for postfix expression parsing", so `tag·to_string` parses as a method call on a variable. That is right — **in expression position**.

`parse_type_path` is also called from type positions, and a type contains no variables. There every `·` is a path separator, and every module path is lower case, so all of them were rejected.

## Measured

Files passing `sigil check`, clean binaries per commit:

| | nihil |
|---|---:|
| `aeadd52` (merge base) | 104/104 |
| `a0180d3` (#59) | 104/104 |
| `8f92d5c` (#61, main HEAD) | **83/104** |
| this PR | 102/104 |

All 21 lost files are this shape — `&vary std·fmt·Formatter<'_>`, `std·fmt·Result`.

## The change

Type positions parse their paths with the heuristic disabled; expression positions keep #61's behaviour untouched. The flag is read-once, so a type that *contains* an expression (a const generic argument, an array length) is expression position again.

This deliberately does **not** revert #61 — that would reintroduce whatever method-call problem #61 set out to fix. It restores the older behaviour only where #61's reasoning does not apply.

The 2 files still failing are middledot module paths in *expression* position (`std·collections·HashMap·new()`), which the pre-#61 code handled only because its first clause was unconditional. Getting those back properly needs a real decision about the grammar rather than a widening, so this PR leaves them.

## Why this targets `main`

`main` and `develop` diverged 8 commits each from `aeadd52`, and they parse the ecosystem very differently:

| Repository | base | `main` | `develop` | `main` + this PR |
|---|---:|---:|---:|---:|
| `nihil` | 104/104 | 83/104 | 104/104 | 102/104 |
| `morgoth` | 50/259 | **257/259** | 42/259 | **257/259** |
| `lucifer` | 18/26 | 21/26 | 18/26 | 21/26 |

`develop` never took #61, so it never had this bug. It is also missing main's `morgoth` fix — **+207 files** — which is the larger prize whenever the branches are reconciled.

## Verification

`jormungandr/tests`, before and after on the same build configuration:

```
Passed: 780   Failed: 15   Skipped: 2
```

Identical both ways — the same 15 fail either way, all environmental (Kafka/AMQP/WebSocket/GraphQL need brokers and the `protocols` feature; the socket, pty and stdin tests need a terminal the container lacks).

Built `--no-default-features --features minimal`, because LLVM 18 is not installed here. **The LLVM and Cranelift backends were not compiled or exercised** — worth a second pair of eyes, though the change is confined to the parser and produces the same AST shape it always did for paths that already parsed.

New test: `spec/22_symbols/P1_020_middledot_module_path_in_type.sg` — parse error before, prints `42` after.

https://claude.ai/code/session_01WsfG6C1ZdnNUscmcfrChVz


paraphilic-ecchymosis:
Closing as superseded by #74.

That merge carries this fix and its regression test (`spec/22_symbols/P1_020_middledot_module_path_in_type.sg`). It has to: the merge takes main's #61 tightening of `parse_type_path`, which is the regression itself, so the fix belongs there rather than as a separate change to `main`.

This PR also targeted `main` from a feature branch, which #69 now rejects — so closing it is the right outcome on both counts.

The diagnosis stands and is preserved in #74: nihil went 104/104 → 83/104 at #61, and every one of those 21 files was a lowercase module path in type position.

---

PR #67, opened 2026-09-09
fix(migrate): rewrite Rust's `core` and `alloc` onto Sigil's `std`

paraphilic-ecchymosis:
## Summary

`core` and `alloc` are Rust's no_std subsets of `std`. **Sigil registers neither** — but it does register `std`, so the three are not equivalent cases:

| root | Sigil registrations |
|---|---|
| `std·` | **31** — `std·fs·read_to_string`, `std·process·exit`, `std·thread·spawn`, … |
| `core·` | 0 |
| `alloc·` | 0 |

Sigil's own ecosystem libraries import `std·collections·`, `std·time·`, `std·sync·`, `std·cmp·`. So `std::` → `std·` was always correct. `core::` and `alloc::` migrate to module roots that don't exist — they parse, and resolve to nothing.

> Correcting my own earlier report: I previously listed `std·`/`core·`/`alloc·` together as ~2,692 sites needing a Sigil stdlib mapping. That was wrong — `std·` is native. The genuinely broken set is 263 sites.

## Two rules

**Prelude types collapse to bare names** — what a Sigil programmer writes, and what Sigil's prelude provides:

```
core·option·Option   →  Option
alloc·string·String  →  String
alloc·vec·Vec        →  Vec
```

The collapse also drops a qualifier in front of the path. Generated code rarely names these directly: prost emits `::prost::alloc::string::String`, its own re-export for no_std builds. **The aether corpus contains no bare `alloc::string::String` at all** — all 160 sites are prost-qualified — so a root-anchored rule would have missed every one of them. I found this only by checking the output rather than trusting the first implementation.

**Everything else moves onto `std`, but only at a path root:**

```
core·sync·atomic·Ordering  →  std·sync·atomic·Ordering
std·alloc·Layout           →  std·alloc·Layout      (unchanged)
```

That second case is the reason `rewrite_path_root` excludes a preceding `·` as well as identifier characters. `std::alloc` is a real `std` path — the engine imports `std::alloc::{alloc, dealloc, Layout}` — and rewriting its second segment would produce `std·std·Layout`. There's a test pinning it.

## Verified end to end

Every merged migration fix, exercised together on one file:

```
//! This crate provides maths for the std::fmt case, in a loop.   ← prose intact
invoke tome·state·Editor;                                          ← #65
☉ Σ S {
    ☉ name: String,                       ← ::prost::alloc::string::String
    ☉ pos: Option<Vec3>,                  ← ::core::option::Option
    ☉ tags: Vec<u8>,                      ← ::prost::alloc::vec::Vec
    ☉ ord: std·sync·atomic·Ordering,      ← core::sync::atomic::Ordering
}
☉ rite f() -> Result<(), gltf·Error> {    ← #64, no leading ·
    ≔ msg = "this crate is not a tome";   ← literal intact
    invoke std·alloc·{alloc, Layout};     ← not std·std
    Ok(())
}
```

Parses successfully. On the real generated protobuf module: **zero `core·`/`alloc·` remaining**.

```
cargo test --release --no-default-features --features jit,native --bin sigil
#   15 passed; 0 failed
```

rustfmt-clean (0 diffs touch the new code). Build verified with `--features jit,native`; LLVM 18 isn't available here, so that configuration is left to CI, which has passed it on #63, #64 and #65.

## Notes

Rebased onto `develop` after #64 and #65 merged; no conflicts. Branched from `develop` and targets `develop` per `CLAUDE.md`.

Remaining Rustisms in the corpus after this, for reference:

| construct | sites | files | |
|---|---|---|---|
| `r#type` | 128 | 60 | raw identifiers; exist only to escape Rust keywords, so each needs a real rename |
| `for<'de>` | 3 | 3 | higher-ranked trait bounds |

Separately, and not a migration matter: Sigil reserves 89 ASCII words as keywords, including prose aliases like `aspect` (= `trait`) and `of` (= `∈`). Those collide with ordinary identifiers — `aspect` alone breaks 14 corpus files, `of` breaks 7 via `TypeId::of::<T>()` — and `location` (247), `body` (164), `from` (101) and `to` (91) are waiting behind them. That's a language design question, not something the migration should paper over by renaming user code.

https://claude.ai/code/session_01KBMCTye5QEFfyXabxvKYrE

---
_Generated by [Claude Code](https://claude.ai/code/session_01KBMCTye5QEFfyXabxvKYrE)_

### DIFF
--- a/parser/src/main.rs
+++ b/parser/src/main.rs
@@ -5187,6 +5187,74 @@ fn drop_absolute_path_prefix(text: &str) -> (String, usize) {
     (out, dropped)
 }
 
+/// Rewrite `from` to `to` where `from` begins a path, returning the new text
+/// and how many roots were rewritten.
+///
+/// "Begins a path" means the preceding character is neither an identifier
+/// character nor `·`. That second exclusion is what keeps `std·alloc·Layout`
+/// intact while rewriting a bare `alloc·Layout`: `std::alloc` is a real `std`
+/// path, and rewriting its second segment would produce `std·std·Layout`.
+fn rewrite_path_root(text: &str, from: &str, to: &str) -> (String, usize) {
+    let mut out = String::with_capacity(text.len());
+    let mut last = 0;
+    let mut rewritten = 0;
+    for (idx, _) in text.match_indices(from) {
+        if idx < last {
+            continue;
+        }
+        let at_root = text[..idx]
+            .chars()
+            .last()
+            .is_none_or(|c| !(c.is_alphanumeric() || c == '_' || c == '·'));
+        if at_root {
+            out.push_str(&text[last..idx]);
+            out.push_str(to);
+            last = idx + from.len();
+            rewritten += 1;
+        }
+    }
+    out.push_str(&text[last..]);
+    (out, rewritten)
+}
+
+/// Collapse a qualified path ending in `suffix` to `to`, dropping any crate
+/// qualifier in front of it. Returns the new text and how many were collapsed.
+///
+/// Generated code rarely names the no_std prelude paths directly: prost emits
+/// `::prost::alloc::string::String`, its own re-export for no_std builds. That
+/// is the same `String`, so the qualifier is noise once the path is no longer
+/// Rust's.
+fn collapse_qualified_path(text: &str, suffix: &str, to: &str) -> (String, usize) {
+    let mut out = String::with_capacity(text.len());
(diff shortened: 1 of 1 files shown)


paraphilic-ecchymosis:
Superseded by #85 — the same change reapplied onto current `develop`, which #68 moved out from under this branch. Your keyword note in the description was right, and is now measured: see #82 and #68.

PR #68, opened 2026-09-09
feat(migrate): warn when a Rust name is reserved in Sigil

paraphilic-ecchymosis:
Warn at migration time when a Rust name will be refused by the Sigil parser once translated, so the reader sees the field rather than a token error later.

## What it does

`sigil migrate` scans the code spans of each file (comments and string literals excluded, per #63) for names in a declaring position, and reports the ones the parser will not accept where they stand:

```
warning: 2 names in this file are refused by the Sigil parser where they stand:
    aspect           2 uses
    simd             1 use
  Rename them in the Rust source, or in the migrated output.
```

## How a name is judged

By **running the parser** on a minimal Sigil program that places the name in the position it was found in, memoised per `(word, position)`. The lexer's token table supplies the candidates — so the list cannot drift from the language — but the parser supplies every verdict.

This is deliberately not a lexer test. **Lexing as a keyword is not the same as failing to parse**: the parser accepts most keyword tokens as names in most positions. Measured over the 93 word-shaped entries in the lexer's table, with an ordinary identifier as a control in every position:

| verdict | count | examples |
|---|---|---|
| accepted everywhere | 58 | `body` `layer` `location` `header` `scope` `anima` |
| refused everywhere | 15 | `aspect` `of` `alter` `each` `vary` `switch` |
| refused as a field, accepted as a binding | 10 | `self` `Self` `this` `true` `yay` |
| accepted as a field, refused as a binding | 10 | `tome` `ref` `super` `const` `simd` |

So a lexer test reports `body`, `layer`, `location` and `header` and tells the reader to rename four names that parse perfectly well.

**A single verdict per word cannot be correct either**, because position decides the answer and does so in both directions:

| word | as a field | as a parameter |
|---|---|---|
| `tome`, `simd` | accepted | refused |
| `this`, `true` | refused | accepted |

Position comes from the innermost unclosed bracket — `(` is a parameter list, `{` a struct body — with `let` / `const` / `static` overriding both. A name whose position cannot be determined is left alone; guessing is the defect this avoids.

Two consequences of the same reasoning:

- Every probe **binds the name and reads it back**. `const`, `async`, `move`, `unsafe`, `volatile`, `atomic` and `simd` bind cleanly and fail only on use, so a probe that binds alone declares them fine.
- The read is placed somewhere **neutral**, never at the start of a statement. A word that introduces an item form (`actor`, `extern`, `invoke`, `rite`, `scroll`, `sigil`, `static`, `type`) is ambiguous with that form when it opens a statement — a property of the use site, not of the name. Reading them there reported eight usable names as refused.

## Corpus

6091 Rust files, from the crate registry and this repository:

| | files | names | wrong |
|---|---|---|---|
| lexer oracle (original) | — | — | reports `body`/`layer`/`location`/`header` |
| parser oracle, before this branch's last commit | 38 | 108 | 72 |
| current | 17 | 44 | none found |

Every one of the 44 was inspected. The 72 that went away were `$macro:ident`-style macro fragments, macro invocation arguments, `const _: () = assert!(…)`, `_` as a discard, and `Self: Sized` following an `FnMut(..)` bound.

Recall improved at the same time: the scanner was restarting its bracket stack at every comment, so a field under its own doc comment had no enclosing brace and was skipped — the ordinary shape of a documented Rust field.

## Tests

24 passing in `migrate_span_tests`, ten of them specific to this:

- a word the parser accepts is not reported (`body`, `location`, `layer`, `header`)
- the same word is judged by where it sits (`tome` and `true`, in both positions)
- a name that only fails when read back is still caught (`const`)
- a word that introduces an item form is not reported, and the seven that genuinely fail still are
- a comment does not lose the enclosing brace
- a macro fragment specifier and a macro argument are not declarations
- a `const` / `static` item binds rather than declaring a field
- a `where` bound is not a declaration, including across an `FnMut(..)` bound
- an undeterminable position stays quiet

`cargo test --release --no-fail-fast` in `parser/`: **564 passed, 3 failed**. The same 3 fail on `develop` at 334fd9d (554 passed, 3 failed) — `llvm_codegen::llvm::tests::test_generic_struct_basic`, `test_generic_struct_two_params`, `parser::tests::test_parse_actor` — and are unrelated to this change.

## Known gaps

- A closure parameter (`|tome: u32|`) is judged as a field, because the enclosing bracket is the function body. No occurrence appeared in the corpus, so this is unmeasured rather than known-harmless.
- `state` is not a keyword; `states` is (`lexer.rs:487`).

## History

The first version asked the **lexer** whether a word produced a keyword token, and the rationale repeated the error in prose, naming `body`, `location`, `layer` and `header` as words that "will not parse" — all four are valid field names. The user-facing warning said "reserved in Sigil and will not parse", which was false for most of what it reported.

Caught in review, and reached independently by #83, which fixes the same false premise in the Python migration tool. Corrected in e25f8b8; the oracle's remaining read-back fault and four position-classifier faults in 3e1899a.


paraphilic-ecchymosis:
## The four names in the doc comment do parse — measured

This PR's rationale says:

> Sigil reserves a number of ordinary English words — `aspect` for `Θ`, `of` for `∈`, and `body`, `location`, `layer`, `header` among others — so Rust code using them as field or binding names migrates into something that **will not parse**

and the warning it emits reads `… reserved in Sigil and will not parse:`.

Tested against a release binary built from `develop`, with an ordinary identifier as control:

```
control:
  foo        PARSES

named by this PR as "will not parse":
  body       PARSES
  location   PARSES
  layer      PARSES
  header     PARSES

correctly identified:
  aspect     fails
  of         fails
```

Each case is `Σ F { NAME: Int }` plus `F { NAME: 1 }` in a `rite main()`. `aspect` and `of` are right; the other four are not.

## The underlying conflation

The classification rests on:

> a word is flagged when lexing it yields anything but the identical identifier

**Lexing as a keyword is not the same as failing to parse.** The parser accepts most keyword tokens as identifiers in most positions. A sweep of all 93 word-shaped lexer tokens across 11 positions, each with an ordinary-identifier control, found:

| behaviour | count | examples |
|---|---:|---|
| refused everywhere | 15 | `aspect alter of each mut vary where loop macro yield switch forever asm` |
| refused only in **name** position | 10 | `self Self this This true false yay yea nay _` |
| refused in the **opposite** direction | 3 | `ref super tome` — accepted as fields, refused as parameters/locals |
| binds, then cannot be read back | 7 | `const async move unsafe volatile atomic simd` |
| **accepted everywhere** | **58** | includes `body layer scope anima` |

So a lexer-token test over-reports by roughly 58 of 93. As written, this warns on names that are completely fine and tells the reader to rename them.

Two structural consequences beyond the false positives:

- The guard flags `name:` — which the comment notes covers "struct fields and typed bindings". But `ref`/`super`/`tome` are accepted as fields and refused as parameters, so those two positions cannot share one verdict.
- `const async move unsafe volatile atomic simd` bind cleanly and fail only when read back. A check that exercises the binding alone will miss them entirely, in the opposite direction.

## Suggestion

The mechanism is worth having — a migration-time diagnostic naming the field rather than the token is a real improvement. It is the *oracle* that is wrong. Rather than "does the lexer return an identifier", ask whether the parser accepts the name **in the position it was found**, e.g. by `sigil check` on a minimal program per position, memoised. #83 does exactly this for the Python migrate tool and reduced its false positives to zero across a 1,529-file corpus.

Same false premise is tracked as #82 (`sigil migrate`'s keyword rename), fixed in #83. Flagging here because this PR would carry it into the Rust codebase, where it becomes the reference implementation rather than a tool-local bug.

paraphilic-ecchymosis:
You're right, and the rationale prose was wrong in exactly the way you describe. Fixed in e25f8b8.

## Reproduced independently

Before changing anything I swept all 111 word-shaped `#[token("…")]` entries from `parser/src/lexer.rs` across three name positions, with an ordinary identifier as a control on each template. It agrees with yours on every overlapping case:

```
             field   param   local
foo (ctrl)   ok      ok      ok
body         ok      ok      ok
location     ok      ok      ok
layer        ok      ok      ok
header       ok      ok      ok
aspect       no      no      no
of           no      no      no
tome         ok      no      no
ref          ok      no      no
super        ok      no      no
true         no      ok      ok
const        ok      no      no
```

**50 of 111 are accepted in all three positions.** The old check reported every one of them.

My token count differs from your 93 — I extracted from the `#[token("…")]` attributes and got 111 — but the grouping matches wherever we overlap, including the three-way split you identified. My sweep also puts `actor`, `extern`, `invoke`, `rite`, `scroll`, `sigil`, `static` and `type` in the "fine as a field, refused as a parameter" bucket alongside `ref`/`super`/`tome`, which makes the positional point stronger rather than weaker.

## What changed

The oracle now runs the parser on a minimal program for the position the name was found in, memoised per `(word, position)`.

One deliberate difference from #83. You noted there that `, name :` is ambiguous and that a wrapped parameter list reads as a struct field, leaving `ref`/`super`/`tome` unrenamed — and that erring toward under-reporting was the right side to fail on. Agreed for a tool that *rewrites* files. But this one only warns, and it turns out the ambiguity is avoidable: tracking the innermost unclosed bracket during the scan distinguishes a parameter list from a struct body exactly, whatever the line breaks. `(` is a parameter, `{` is a field, `let`/`let mut` overrides both. Those three keywords come out right.

Your other two structural points are both addressed:

- **`name:` cannot share one verdict across positions** — it no longer does; position is resolved per occurrence.
- **Binding position alone misses the `const` family** — the local probe binds *and* reads back (`≔ const = 1; const + 1`), so `const`, `async`, `move`, `unsafe`, `volatile`, `atomic` and `simd` are caught.

A name whose position can't be determined is now left alone rather than guessed at.

Also worth noting: this calls `sigil_parser::Parser` in-process rather than shelling out to `sigil check`, so there's no binary to locate and no path where a missing one turns the pass into a silent no-op.

## Tests

19 passing, four new and one per failure mode you named: a word the parser accepts is not reported; the same word judged by where it sits (`tome` and `true`, both positions); a name that only fails on read-back; an undeterminable position staying quiet.

Thanks for catching this before it became the reference implementation — that was the right call, and the note in #83 flagging it here is what surfaced it.

---

paraphilic-ecchymosis:
## Corpus measurement: 759 flags → 24

Both oracles run over the same 873 Rust files (the `aether` engine, `engine/**/*.rs`), old binary built from c201e3b, new from ace4aa4.

| | files warned | flags | distinct names |
|---|---:|---:|---:|
| old (lexer test) | 186 | 759 | 23 |
| new (parser, per position) | 13 | 24 | 2 |

The old oracle's top entries are all names that parse perfectly well:

```
location  244      to        89       states    39
from       98      layer     50       linear    17
body       93      trigger   42       timeout   12
```

`location` alone accounts for a third of the output, and every one of those 244 was a false positive telling the reader to rename a field that was already fine.

What survives is `aspect` (21) and `actor` (3). Both check out by hand:

- **`aspect`** — `pub aspect: f32` in `aether-render/src/camera.rs` and `aether-gpu/src/texture.rs`, plus `fn perspective(fov_degrees: f32, aspect: f32, …)`. Refused in every position.
- **`actor`** — three parameters in `aether-physics/src/game_systems/emergence/`, each in a parameter list wrapped across lines:

  ```rust
  pub fn record_event(
      &mut self,
      actor: EvolutionActor,
  ```

  This is the case you flagged in #83 as one the line-shape heuristic reads as a struct field, leaving `ref`/`super`/`tome` unrenamed. `actor` is the same shape of problem — valid as a field, refused as a parameter — and reading position from the innermost unclosed bracket rather than the line gets all three right.

## One more false positive, found by this run

The first corpus pass flagged 25, not 24. The extra was `Self` in `aether-core/src/ecs.rs`:

```rust
fn query(world: &World) -> QueryIter<'_, Self>
where
    Self: Sized;
```

A `where` bound puts a name in front of a colon inside a trait body — the same shape as a field declaration, and `Self` is not a valid field name. Nothing is declared there.

Fixed in ace4aa4 by tracking the clause from the `where` keyword to the body or semicolon that ends it, rather than matching the keyword alone, so the comma-separated form `where T: A, U: B` is covered too. 20 tests pass.

Worth saying plainly: all 19 unit tests passed with that false positive present. It took running against real source to find, which is an argument for the corpus check being part of this rather than a one-off — happy to add it as a test fixture if you'd like it standing.

---

Issue #73, opened 2026-09-09
`run-dir` prints "Crate name:" — a Rustism in user-facing CLI output


paraphilic-ecchymosis:
## Summary

`sigil run-dir` labels the module group it loaded as a "Crate":

```
$ sigil run-dir .
Loading 1 modules from '.':
  - main.sg
Crate name: my_project
```

`crate` is Rust's word. Sigil's own term for the unit is `tome` — that's what the language reserves, and what the migration tool now translates Rust's `crate` into (#65).

Having the CLI print the Rustism back at the user undercuts that: someone reading `run-dir` output learns the wrong noun for the concept, and it's the noun they'll then reach for in their own code, where it's a reserved word that means something else.

## Suggested change

```
Tome name: my_project
```

Worth a sweep of the rest of the user-facing strings at the same time — this is likely not the only place the Rust vocabulary leaked into output.

paraphilic-ecchymosis:
Fixed by #88, merged into `develop` as 4b07854. All nine user-facing occurrences in `main.rs` swept to "tome", not just the one string; the `[LLVM]`/`[DEBUG]`/`[IMPL-REG]` diagnostics keep "crate" since they name internal structures rather than user vocabulary.

(Closing by hand — the "Closes" keyword only fires on merge into the default branch `main`, and we merge into `develop`.)

Issue #72, opened 2026-09-09
`run-dir` calls `main` with one argument, so every zero-arg `main` errors after running


paraphilic-ecchymosis:
## Summary

`sigil run-dir` passes an argument to `main`, so a program whose `main` takes no parameters runs to completion and *then* fails. `sigil run` on the same file is clean.

## Repro

`main.sg`, nothing else in the directory:

```sigil
☉ rite main() {
    println("hello");
}
```

## Actual

```
$ sigil run main.sg
hello

$ sigil run-dir .
Loading 1 modules from '.':
  - main.sg
Crate name: tmp.JvPCp6UvqX
hello
Runtime error: [R0000] Runtime error: Expected 0 arguments, got 1 (func=Some("main"), params=[])
```

Note the body executes (`hello` prints) — the arity check fires on the call itself, so the failure is purely spurious. But it sets a non-zero exit and prints an error, which makes `run-dir` unusable as a "does this crate work" signal: you can't distinguish a real failure from this one without reading the message.

## Expected

```
hello
```

with a zero exit status, matching `sigil run`.

## Fix sketch

`run` and `run-dir` should agree on how they invoke the entry point. Either drop the extra argument in the `run-dir` path, or accept both `main()` and `main(args)` arities in both paths.

paraphilic-ecchymosis:
Fixed by #88, merged into `develop` as 4b07854.

(The PR said "Closes #72" but that keyword only fires on merge into the repo default branch, `main`. Everything here merges into `develop` first, so it never fires — closing by hand.)

PR #75, opened 2026-09-09
fix(interpreter): dispatch operator impls and evaluate associated consts

### LINKED ISSUE #70: Interpreter does not dispatch user operator-trait impls (`⊢ Add ∀ T` + `a + b` → "Invalid struct operation")
## Summary

A user-written operator trait impl (`⊢ Add ∀ T`) type-checks cleanly but is never dispatched at runtime. Applying `+` to two struct values fails with `Runtime error: Invalid struct operation` instead of calling the impl's `add`.

This is a checker/interpreter divergence: `sigil check` accepts the program, `sigil run` rejects it.

## Repro

```sigil
☉ Σ P { ☉ x: f32 }

⊢ P {
    ☉ const ONE: Self = Self { x: 1.0 };
    ☉ rite new(x: f32) -> Self { Self { x } }
}

⊢ Add ∀ P {
    type Output = Self;
    rite add(self, o: Self) -> Self { P·new(self.x + o.x) }
}

☉ rite main() {
    ≔ a = P·new(2.0);
    ≔ b = a + a;              // user Add impl
    println(to_string(b.x));
}
```

## Actual

```
$ sigil check ops.sg
✓ ops.sg - no errors

$ sigil run ops.sg
Runtime error in 'ops.sg': [R0000] Runtime error: Invalid struct operation
```

## Expected

```
4
```

## Notes

- Not specific to `Add`. A `Mul<f32>` impl on the same shape reproduces identically, and also type-checks clean.
- Calling the method directly (`P·add(a, a)` / `a.add(a)`) works — only the operator sugar fails to route to the impl, which points at operator evaluation in the interpreter rather than at trait resolution.
- The type checker's acceptance suggests trait resolution already knows about the impl; the interpreter's binary-op path appears to fall through to a built-in struct-operation branch without consulting user impls.

## Impact

This blocks running any Sigil code that uses operator overloading. Concretely, it stops the `aether-sigil` math crate from executing its own test suite — `Vec2`/`Vec3`/`Vec4` define `Add`/`Sub`/`Mul`/`Div`/`Neg` impls, and every arithmetic expression in that crate hits this path.

### LINKED ISSUE #71: Associated consts initialised by a call don't evaluate ("no field 'x' on struct 'Self'")
## Summary

An associated const initialised with a struct literal evaluates correctly. The same const initialised by calling an associated `rite` fails at runtime with `no field '<field>' on struct 'Self'` — the constructor appears to return an unresolved `Self` rather than an instance.

`sigil check` accepts both forms.

## Repro

```sigil
☉ Σ P { ☉ x: f32 }
⊢ P {
    ☉ const rite new(x: f32) -> Self { Self { x } }
    ☉ const LIT: Self = Self { x: 1.0 };
    ☉ const CALL: Self = Self·new(2.0);
}
☉ rite main() {
    println("literal init : " + to_string(P·LIT.x));
    println("call init    : " + to_string(P·CALL.x));
}
```

## Actual

```
$ sigil check c1.sg
✓ c1.sg - no errors

$ sigil run c1.sg
literal init : 1
Runtime error in 'c1.sg': [R0000] Runtime error: no field 'x' on struct 'Self'
```

## Expected

```
literal init : 1
call init    : 2
```

## Notes

Things ruled out while narrowing this down:

- **Not** associated consts in general — the literal form on the line above works.
- **Not** cross-module resolution — the repro is a single file, and consts resolve fine across modules when literal-initialised.
- **Not** the `const rite` marker specifically — writing the initialiser as `P·new(2.0)` instead of `Self·new(2.0)` fails the same way.

The distinguishing factor is the *form* of the initialiser: struct literal works, call expression does not. The error text naming the type as `Self` (rather than `P`) suggests the const initialiser is evaluated in a context where `Self` hasn't been bound to the concrete type, so the call returns something whose fields can't be projected.

## Impact

This is the second blocker (with #70) stopping the `aether-sigil` math crate from running. `Vec3·X`, `Vec3·Y`, `Vec3·Z`, `Vec3·ZERO`, `Vec3·ONE` and their `Vec2`/`Vec4` counterparts are all declared as `☉ const X: Self = Self·new(1.0, 0.0, 0.0);` — the call form — so every named vector constant in the crate hits this.

paraphilic-ecchymosis:
Fixes #70. Fixes #71.

Three defects that between them stopped any Sigil program using operator overloading or named constants from running — all three accepted by `sigil check`, all three failing only at runtime.

## 1. Operators never reached user impls (#70)

`eval_binary` matched operand types against a table of built-ins and, finding nothing, reported `Invalid struct operation`. It never looked for an `impl Add for T`. The unary path had the same gap.

Both now consult the user's impls when the built-in table has nothing to offer, so built-ins keep priority and the arithmetic hot path is untouched — the operands are only cloned when one side is a struct.

Dispatch needs its own registry rather than the existing `Type·method` globals, because that name cannot tell two impls of one operator apart:

```sigil
⊢ Div<f32> ∀ Vec3 { rite div(self, s: f32) -> Self { ... } }
⊢ Div<Vec3> ∀ Vec3 { rite div(self, o: Self) -> Self { ... } }
```

Both define `Vec3·div`. The second overwrites the first in `globals` and the other becomes unreachable — silently. Impls are now filed under `(self type, trait)` with their declared right-hand type kept alongside, and `/` picks the one whose operand type matches. A type with a single impl of a trait skips the check entirely; when nothing matches exactly the first impl still runs, so the error surfaces inside the method where it says something useful.

This also covers impls on primitives (`⊢ Mul<Vec3> ∀ f32`), which are reached by looking up the left operand's type name.

## 2. Associated consts initialised by a call did not evaluate (#71)

Two causes, both needed fixing:

- **Order.** Impl items were processed top to bottom, so a const calling a constructor declared further down ran before that constructor existed. This is the common layout — a type's constants are conventionally written above its `new`.
- **`Self` unbound.** `current_self_type` was not set while const initialisers ran, so `Self·new(1.0, 0.0, 0.0)` resolved to nothing and fell into the unknown-type path, which quietly returns an **empty struct named `Self`**. That is where the misleading `no field 'x' on struct 'Self'` came from — the failure was in the constructor call, not the field access.

Methods are now registered in a first pass and consts evaluated in a second, with `Self` bound to the impl's type.

The same unbound `Self` broke dispatched operator calls, whose bodies are typically `Self·new(self.x + o.x)` — so the dispatch in part 1 binds it too.

## 3. Impls inside a module dropped their consts entirely

Found while verifying the above, and not previously filed. The module path registered methods and ignored every `ImplItem::Const`:

```sigil
scroll geom {
    ☉ Σ P { ☉ x: f32 }
    ⊢ P {
        ☉ const rite new(x: f32) -> Self { Self { x } }
        ☉ const ONE: Self = Self·new(1.0);
    }
}
// before: Runtime error: Undefined: geom·P·ONE (tried geom·P·ONE and ONE)
```

Fixing only the top-level path would have shipped a fix that silently does not apply inside a module — a worse trap than the original bug, since the code would look correct. That path now evaluates consts as well, and registers operator impls.

## Verification

`cargo test --no-default-features --features minimal,protocols`: **423 passed, 0 failed**, including 5 new regression tests — one per defect, plus a guard that integer and string operators are unaffected.

Behaviour was checked against the vector module of the `aether` Sigil port, whose `Vec2`/`Vec3`/`Vec4` exercise every part of this. Each result matches the Rust implementation it was ported from:

| expression | result |
|---|---|
| `(1,2,3) + (4,5,6)` | `5 7 9` |
| `(1,2,3) - (4,5,6)` | `-3 -3 -3` |
| `(1,2,3) * 2.0` | `2 4 6` |
| `(4,6,8) / 2.0` | `2 3 4` |
| `(4,6,8) / (2,3,4)` | `2 2 2` ← the shadowed `Div<Vec3>` |
| `-(1,2,3)` | `-1 -2 -3` |
| `2.0 * (1,2,3)` | `2` ← impl on a primitive |
| `Vec3·X.cross(Vec3·Y)` | `0 0 1` ← the constants |
| `(3,4,0).normalize().length()` | `1` |
| `(1,2,3).lerp((4,5,6), 0.5)` | `2.5 3.5 4.5` |
| `(2,3,0).project_onto(Vec3·X)` | `2 0 0` |

None of these ran before this change.

## Not included

#72 (`run-dir` calls `main` with one argument) and #73 (`run-dir` prints "Crate name:") are both in `main.rs` rather than the interpreter, so they belong in their own change rather than mixed in here.

https://claude.ai/code/session_01KBMCTye5QEFfyXabxvKYrE


paraphilic-ecchymosis:
## Merged `develop` in; conflicts resolved

`develop` had moved this file by +3058/-149 under the branch (#62, #74, #85). Merge commit: `6d71e0b`. Two conflict regions, both in `parser/src/interpreter.rs`.

### Operator dispatch — kept wholesale

`develop` has zero references to `operator_impls` or `is_operator_trait`; nothing upstream duplicates this work, so the conflict was positional rather than semantic. All of it survives, verified after the merge: the field and its init, both registration sites (top-level and in-module), `is_operator_trait` / `operator_trait_method` / `binop_trait` / `unaryop_trait`, `lookup_operator_impl`, `try_binary_overload`, `try_unary_overload`, `call_operator_impl`, and the `overload_operands` wiring in `eval_binary` / `eval_unary`.

**Verified working, not taken on trust.** The same source run against plain `origin/develop`:

```
impl Add for V3 { ... }   a + a
develop:  Err("Invalid struct operation")
merged:   Ok(2.0)         # correct
```

New test `operator_dispatch_runs_a_user_add_impl_end_to_end` exercises the real shape — a three-field vector with `Add`, `Sub` and `Mul<f32>` chained in one expression, `(a + b - b) * 2.0`.

### Associated consts — this branch's version, and there was no real choice to make

Both sides evolved this path, but they are not rivals: **this branch's rewrite is a strict superset of develop's.**

Develop's version (from #74) does two things. It sets `current_self_type` around the initialiser so `Self` resolves, and it defers a const whose initialiser mentions a const-generic param. This branch's `register_impl_const` does **both, identically** — it had independently reached the same `current_self_type` fix — and neither const generics nor deferred evaluation is dropped. It then adds two things develop does not have:

1. **Item ordering.** Methods are registered before consts, so `☉ const ONE: Self = Self·new(1.0)` works when `new` is declared *after* it. Impl items are unordered; develop's single in-order pass fails this.
2. **Consts in module impls.** Develop's `scroll` path handles only `ImplItem::Function` and drops every const silently — `P·ONE.x` reads back `null` rather than erroring. Measured on develop: `Ok(null)`; on the merge: `Ok(1.0)`.

So no capability is lost in either direction and nothing needed escalating to the original author. Develop's richer `"Type mismatch in binary operation"` diagnostic is also kept, as the fallback arm, with the overload retry after it.

### One thing found on the way, unrelated to this branch

`module·Type·member` path resolution is **broken on develop**. It resolves at the merge base and fails on develop with ``undefined variable: `geom` `` — for a plain static method, no consts and no operator traits involved:

```
scroll geom { ☉ Σ P { ☉ x: f32 }  ⊢ P { ☉ rite new(x: f32) -> Self { Self { x } } } }
rite main() { ⤺ geom·P·new(3.0).x; }

merge base (7adc3af):  Ok(3)
origin/develop:        Err("undefined variable: `geom`")
```

Something between `7adc3af` and `42ee1f0` broke it. It is out of scope here and left for its own change. Its only effect on this PR is that `impl_inside_a_module_keeps_its_associated_consts` now asserts through the unqualified `P·ONE` alias instead of `geom·P·ONE` — which still distinguishes the fix cleanly (`1.0` vs develop's `null`). `module_qualified_paths_are_unresolvable` is an `#[ignore]`d record of the regression; un-ignore it when it is fixed.

### Tests

`cargo test --release --no-fail-fast` in `parser/`, both measured locally on this machine:

| | lib | bin |
|---|---|---|
| `origin/develop` (baseline) | 551 passed, 2 failed | 25 passed |
| this branch, merged | **557 passed, 2 failed, 1 ignored** | 25 passed |

The 2 failures are identical on both sides and pre-existing: `llvm_codegen::llvm::tests::test_generic_struct_basic` and `test_generic_struct_two_params`. No regressions; +6 passing tests.

PR #78, opened 2026-09-10
release: merge develop into main (closes the divergence)

paraphilic-ecchymosis:
Closes the `main`/`develop` divergence rather than moving it. Follows #74.

## Why this is needed

`main` has been at **nihil 83/104** since #61 tightened `parse_type_path`. #74 reconciled the branches on `develop` (restoring 102/104 while keeping main's morgoth fix, 257/259 vs develop's pre-merge 42/259) — but the release-merge back to `main` is the half nobody had done. Until it lands, the divergence has been *moved*, not closed.

## Why the tree is taken wholesale from develop

**#74 was squash-merged**, so `main` is not an ancestor of `develop`. Git recomputes the merge base at the original divergence point and re-raises conflicts in the same five files #74's 97 hunks already settled: `interpreter.rs`, `lexer.rs`, `llvm_codegen.rs`, `main.rs`, `parser.rs`.

Re-resolving those by hand would risk silently contradicting decisions made with measurements behind them. So this merge takes develop's tree wholesale — verified, not assumed:

- **No file exists on `main` that `develop` lacks.**
- The real delta is **10 files, +536/−56** (GitHub's larger figure is an artifact of the broken ancestry).
- The 56 deletions are intentional supersessions: `For`/`In`/`Break`/`Continue` are dead lexer tokens with no rule producing them on either branch, and the rest is `main.rs` migrate logic that develop's #63/#64/#65 rewrote.
- **The resulting tree is byte-identical to `origin/develop`.**

The merge commit has both parents, so it also **repairs ancestry** — future `develop` → `main` merges will behave normally instead of re-conflicting.

## Reviewer note

Worth confirming the wholesale-resolution call, since it is the one judgement here. Everything else is mechanical.

Separately, `develop` carries five `:Zone.Identifier` files (25-byte Windows NTFS artifacts under `kit/methodologies/`) that this merge carries to `main`. Left alone deliberately so this stays a verifiable "main becomes develop" — worth deleting in a follow-up.


paraphilic-ecchymosis:
Closing in favour of #77, which was opened three minutes earlier and solves this better.

We diagnosed the same root cause independently — #74 landed as a squash, so `main` is not an ancestor of `develop` and the release merge re-conflicts in the same five parser files #74's 97 hunks already settled.

#77 is the better fix. It records the missing ancestry with an `-s ours` merge and changes **zero files**, after which the release merge is a plain fast-forward. This PR instead resolved those five conflicts by taking develop's tree wholesale — which I verified as safe (byte-identical to `origin/develop`, no main-only files, the 56 deletions all intentional supersessions), but it is still a judgement call baked into a 10-file merge. #77 removes the need to make that call at all.

One thing worth carrying over: `develop` currently has five `:Zone.Identifier` files (25-byte Windows NTFS artifacts under `kit/methodologies/`) that a release merge will carry to `main`. Worth deleting separately.

PR #80, opened 2026-09-10
merge: resolve #62 against develop (reviewable on its own)

paraphilic-ecchymosis:
Resolves #62's conflict with `develop`. Opened as its own PR rather than pushed onto the branch, so the resolution can be read separately instead of disappearing into 98 commits — the same reasoning qliphoth#10's author gave for leaving their conflict unresolved.

## Most of the 310 hunks were phantom

`git merge origin/develop` reported 31 files / ~310 hunks. Most were an artifact of a stale merge base.

**All 27 files** under `parser/src/migrate/react/` and `parser/src/wasm/` on `develop` are **byte-identical to the blob at `a0180d3`** — a commit that is itself an ancestor of this branch. That work travelled from here to `main` by another route and re-entered `develop` via #74. Develop's side in those clusters is a strict ancestor snapshot, so taking the branch side cannot lose anything. Verified by blob hash, not assumed.

Re-basing the 3-way merge on `a0180d3` collapsed the core-parser cluster from 59 hunks to **20 genuine conflicts**. #77 landing mid-work independently confirmed the base was sound.

## Resolutions, by cluster (one commit each)

| Cluster | Took | Why |
|---|---|---|
| `migrate/react/*` (8 files) | branch | develop's copies are this branch's own `a0180d3` blobs |
| `wasm/*` (11 files) | branch | same; taking develop would revert two defect fixes the probe exists to record |
| `parser.rs`, `stdlib.rs` | branch | same `·`-in-type-position fix, explicit parameter over hidden parser state; also handles `tome`/`super` roots |
| `main.rs` | **develop** | develop's code-span scanner handles nested block comments, arbitrary-hash raw strings and byte literals; the branch's has no byte-literal or block-comment handling at all |
| `lexer.rs` | develop | `For`/`In`/`Break`/`Continue` removal, per #74 |
| fixtures (8 files) | branch | develop never touched them; verified no `.sg`/`.expected` pair is split |

## Two deliberate departures

**`llvm_codegen.rs`** — took develop's new work (`sigil_pi`, `Expr::Deref` assign) but removed two things not among #74's recorded decisions:
- a hoisted `idx` at line 11514, shadowed by an identical call at 11523 with an `Expr::Range` early return between them — so `arr[next()]` compiled the index expression **twice**
- a verbatim-duplicated `PassBuilderOptions` block

Worth fixing on `develop` directly, or this conflict recurs on every future merge.

**Outside the merge:** `report_stubbed_calls` took `&WasmCompiler` with no `#[cfg(feature = "wasm")]`, so **this branch did not build under its own default features**. Its identical sibling was already gated; this one was missed.

## Build and tests

All three configurations build: default (LLVM 18), `--features wasm`, `--no-default-features --features jit,native`.

`cargo test --release --no-fail-fast`: **551 passed / 2 failed** (lib), 14 passed (bin). The two failures — `test_generic_struct_basic`, `test_generic_struct_two_params` — were verified to fail on **the branch alone and on develop alone**. Pre-existing, not merge-induced.

## A gap found while doing this — S12

A resolution was made, tested, found wrong, and reversed; both commits are left in place rather than rewritten. The claim was that develop's `CONCLAVE.sigil` does not parse because `anima` is a keyword. It does parse — only `aspect` and `alter` are rejected.

That produced **S12**: `sigil migrate`'s keyword-identifier pass renames on the premise that a keyword can never legitimately be an identifier, which is false. Its visible damage is that running migrate over this repo rewrote **nine existing entries** in a file whose header reads "APPEND-ONLY: Never modify existing entries."

For a probe whose deliverable is the gaps, that is arguably the more valuable output here.

## Worth a second look

`main.rs` is the largest place this overrides the PR's own work. The capability argument is evidenced but the branch's scanner was never demonstrated to fail on the Qliphoth corpus specifically.


paraphilic-ecchymosis:
## Correction: the stated reason for the `main.rs` resolution is wrong

The merge itself is fine — see the verification below — but the justification recorded in the PR description is false, and a future reader would otherwise trust it.

**What the description claims:**

> `main.rs` | **develop** | develop's code-span scanner handles nested block comments, arbitrary-hash raw strings and byte literals; the branch's has no byte-literal or block-comment handling at all

**That is not true.** The two `map_rust_code_spans` implementations are **byte-for-byte identical**. Extracting both function bodies and diffing them produces no output. They carry the same capability comments, on both sides:

```
// Line comment: // ... including /// and //!
// Block comment: /* ... */, which nests in Rust
// Raw string: r"..", r#".."#, br#".."# and any hash depth
```

The branch's scanner tracks block-comment nesting depth and handles `r"`, `r#"`, `br#"` at arbitrary hash depth, exactly as develop's does. There was no capability gap to resolve in that direction.

## What actually happened, which is the good news

The resolution did **not** replace one side with the other. `parser/src/main.rs` line counts:

| | lines |
|---|---:|
| branch before merge (`25eeaeb`) | 6,807 |
| develop | 6,830 |
| **after this merge** | **7,046** |

Larger than either side, i.e. a union. All seven of the branch's WASM-only functions survived: `default_wasm_output`, `parse_feature_flags`, `report_stubbed_calls`, `report_unresolved`, `report_invalid_module`, `set_wasm_cfg`, `validate_wasm_module`.

So no work was dropped. The outcome is right; the recorded reasoning for it is not.

## Also worth recording

The constructs named in the original justification are nearly absent from the corpus this migration targets. Across all `.rs` files in `Daemoniorum-LLC/qliphoth`: **1** file containing `r#"`, **1** containing `b'`, and **0** containing nested block comments. Even had there been a capability gap, its practical reach would have been two files.

Separately, the PR description reports "551 passed / 2 failed" with the failures verified pre-existing. On unmodified `develop` the baseline is **540 passed / 3 failed** — `test_parse_actor` also fails there. The counts differ because they were measured on different trees; the "pre-existing, not merge-induced" conclusion still holds for all three.

## Provenance of this correction

I reviewed and merged this PR, and I initially confirmed the scanner claim using grep hit-counts rather than reading the function — reporting that develop had 25 byte-literal matches to the branch's zero. That count was an artefact of shell quoting; the branch is full of `b'/'` and `b'*'` byte comparisons. Reading the two functions takes one command and gives the opposite answer.

No action needed on the merge. The `main.rs` conflicts could be re-derived now that the scanners are known identical, but nothing is blocked on it — #62 remains draft.

PR #87, opened 2026-09-10
fix(parser): a module root is lowercase, and so is a receiver

### LINKED ISSUE #86: Regression: module-qualified paths (`module·Type·member`) are unresolvable on develop
A module-qualified path fails at runtime on `develop`. Found while resolving #75, but unrelated to it.

## Reproduction

```sigil
scroll geom {
    ☉ Σ P { ☉ x: f32 }
    ⊢ P { ☉ rite new(x: f32) -> Self { Self { x } } }
}
rite main() { ⤺ geom·P·new(3.0).x; }
```

```
develop (42ee1f0):  [R0003] Runtime error: undefined variable: `geom`
```

**Control, same file, only the qualifier removed:**

```sigil
rite main() { ⤺ P·new(3.0).x; }     →  3
```

So the type and the static method both resolve fine. It is specifically the `module·` qualifier that fails. No consts, no operator traits, no generics involved — a plain static method.

## Scope

Reported to work at merge base `7adc3af` and fail at `42ee1f0`. Every `interpreter.rs` change in that range arrived with #62's branch, authored on or before 2026-09-08.

**Not caused by #62's conflict resolution.** That commit (`2fcd85a`) touches `interpreter.rs` by exactly 2 lines, neither module-related. Ruled out explicitly because the resolution is recent and the obvious suspect.

Candidates still to bisect, all pre-existing on the branch:

| commit | note |
|---|---|
| `c7e003d` | titled "Ph1: to_bool, **module scope**, Self in actors, None as a const, min/max" — the name makes it the first place to look |
| `7580314` | "The higher-order morphemes did nothing, and a sweep to find the rest" |
| `2a0a2ef` | "interpreter: add String slice and find" |

A bisect needs a build per step; note that the #62 branch **did not compile under its own default features** before `3f64f38` (`report_stubbed_calls` took `&WasmCompiler` with no `#[cfg(feature = "wasm")]`), so points before that need `--features wasm` or the gate applied.

## Test already written

#75 carries an `#[ignore]`d regression record:

```
parser/src/interpreter.rs :: module_qualified_paths_are_unresolvable
```

Un-ignore it when this is fixed. #75 also had to route its module-const test through the unqualified `P·ONE` alias to work around this — that workaround can go at the same time.

paraphilic-ecchymosis:
## What

`geom·P·new(3.0)` failed at runtime with ``undefined variable: `geom` `` while `P·new(3.0)` returned 3 — the type and the static method both resolved, and only the `module·` qualifier did not.

It is a **parse** defect, not an interpreter one. `dump-ir` says so outright:

| | |
|---|---|
| working | `call geom::P::new` |
| broken | method call `new` on field `P` of a variable `geom` that was never bound |

## Which commit, and what it actually did

**`8f92d5c` (#61)** — bisected over 131 commits, `01f4d6a` good. That is *much older* than the range #86 proposed, and none of the three candidates (`c7e003d`, `7580314`, `2a0a2ef`) touch path resolution at all; their `interpreter.rs` hunks are `to_bool`, `Set::contains_key` and `String::slice`/`find`. `8f92d5c` predates #62's branch and reached develop only when that branch merged — which is why develop's own tip at `7adc3af` still worked. The issue's "works at `7adc3af`" claim is correct, but it worked *by accident*:

```rust
let is_path_sep = self.consume_if(&Token::MiddleDot)
    || (first_segment_is_type && self.consume_if(&Token::MiddleDot));
```

The first operand always fires, so `first_segment_is_type` never decided anything and every `·` was swallowed as a path separator. #61 fixed a real bug with that — `tag·to_string()` was parsed as a call to a path `tag·to_string` and dispatched to a built-in of the wrong arity — by making the guard load-bearing:

```rust
let is_path_sep = first_segment_is_type && self.consume_if(&Token::MiddleDot);
```

Correct for a receiver, wrong for a module. Both are lowercase, and `first_segment_is_type` cannot tell them apart, so the fix took module-qualified paths down with the misparse. `8e4e326` recorded the survivor as a follow-up — *"`std·collections·HashMap·new()` still fails at runtime with undefined variable: std … it cannot be fixed by enumerating keywords, since any module name may be lowercase"* — and added `tome`/`super` as keyword roots, which covers the two spellings that are keywords and none of the ones that are not.

## The fix

What separates a module root from a receiver is not the root but what the chain leads **to**. A module chain reaches a segment naming a type, because Sigil types are uppercase and methods are not: `geom·P·new` reaches `P`; `tag·to_string` reaches nothing.

So in expression position `·` is a path separator when the `·ident` chain ahead of it reaches an uppercase segment, and once one `·` is accepted the rest of the chain is a path too. The scan stops at the first token that is not `·ident`, so a call, an index or an operator ends it — `xs·map(f)` never looks past the `(`.

Both halves of #61's decision stay true, and the follow-up `8e4e326` recorded falls out as the same rule one segment further along:

```sigil
geom·P·new(3.0).x        → 3     (was: undefined variable: `geom`)
outer·inner·Q·new(7).n   → 7     (was: undefined variable: `outer`)
P·new(3.0).x             → 3     (unchanged)
tag·to_string()          → method call on `tag`   (unchanged)
xs·len(), m·get(k)       → unchanged
```

**The interpreter is untouched.** It has registered `geom·P·new` all along; nothing ever asked it for that name.

## Tests

- `module_qualified_paths_are_unresolvable` was the `#[ignore]`d record of this bug. Un-ignored and renamed `module_qualified_paths_resolve`.
- `a_lowercase_receiver_is_still_a_method_call_not_a_path` pins the other half, so the next person to touch this line has both cases in front of them. #61's own `test_middledot_method_dispatch` still passes.
- `impl_inside_a_module_keeps_its_associated_consts` asserted through the unqualified `P·ONE` alias because #75 could not spell the qualified form. That workaround is gone; it now asserts `geom·P·ONE.x + P·ONE.x`, pinning both.

## Verification

Measured on this machine rather than quoted, baseline taken at `2e6aac6`:

| | baseline | this PR |
|---|---|---|
| lib | 557 pass / 2 fail / 1 ignored | **559 pass / 2 fail / 0 ignored** |
| bins | 25 / 25 | 25 / 25 |
| Sigil suite | 794 pass / 4 fail / 2 skip | 794 pass / 4 fail / 2 skip |

The same two lib failures on both sides — `test_generic_struct_basic` and `test_generic_struct_two_params`, both pre-existing LLVM generic-struct, neither touched here. Sigil-suite per-test results are byte-identical between the two runs.

`sigil check`, `jit` and `llvm` on the reproduction behave exactly as they do on `develop` (the LLVM backend returns a garbage f32 on both sides, and on `7adc3af` too — separate, pre-existing, not this).

Closes #86

### DIFF
--- a/parser/src/interpreter.rs
+++ b/parser/src/interpreter.rs
@@ -26543,13 +26543,13 @@ mod tests {
     fn impl_inside_a_module_keeps_its_associated_consts() {
         // Module impls registered their methods but dropped every const, so
         // `P·ONE` was not merely wrong but absent: reading `.x` off it yielded
-        // null rather than 1.0. (Still the behaviour on develop today.)
+        // null rather than 1.0.
         //
-        // The const is asserted through the unqualified `P·ONE` alias rather
-        // than `geom·P·ONE` because module-qualified path resolution is
-        // currently broken on develop for *everything* — `geom·P·new(..)`, an
-        // ordinary static method that predates this branch, fails the same way.
-        // See `module_qualified_paths_are_unresolvable`.
+        // Asserted through the module-qualified `geom·P·ONE`, which is how the
+        // const is actually spelled from outside the scroll. #75 had to route
+        // this through the unqualified `P·ONE` alias, because the qualified
+        // form could not be parsed at all; see
+        // `module_qualified_paths_resolve`. Both spellings are pinned here.
         let source = "\
             scroll geom {
                 ☉ Σ P { ☉ x: f32 }
@@ -26558,19 +26558,16 @@ mod tests {
                     ☉ rite new(x: f32) -> Self { Self { x } }
                 }
             }
-            rite main() { ⤺ P·ONE.x; }";
-        assert!(matches!(run(source), Ok(Value::Float(v)) if v == 1.0));
+            rite main() { ⤺ geom·P·ONE.x + P·ONE.x; }";
+        assert!(matches!(run(source), Ok(Value::Float(v)) if v == 2.0));
     }
 
     #[test]
-    #[ignore = "pre-existing develop regression, unrelated to operator dispatch"]
-    fn module_qualified_paths_are_unresolvable() {
-        // Regression record, not a claim about this branch. `module·Type·member`
-        // resolves on the merge base and fails on develop with
-        // "undefined variable: `geom`" — for a plain static method, with no
-        // consts and no operator traits involved. Whatever landed in develop
-        // between 7adc3af and 42ee1f0 broke it; fixing that belongs in its own
-        // change. Un-ignore this test when it is fixed.
+    fn module_qualified_paths_resolve() {
+        // `module·Type·member` used to fail with "undefined variable: `geom`":
+        // `parse_type_path` stopped at the first `·` after a lowercase segment,
+        // so the path became a field access on a variable that was never bound.
+        // A plain static method, no consts and no operator traits involved.
         let source = "\
             scroll geom {
                 ☉ Σ P { ☉ x: f32 }
@@ -26578,6 +26575,29 @@ mod tests {
             }
             rite main() { ⤺ geom·P·new(3.0).x; }";
         assert!(matches!(run(source), Ok(Value::Float(v)) if v == 3.0));
+    }
+
+    #[test]
+    fn a_lowercase_receiver_is_still_a_method_call_not_a_path() {
+        // The other half of the same decision, and the reason the qualified
+        // form was broken in the first place: #61 stopped `·` from being
+        // consumed as a path separator after a lowercase segment, because
+        // `tag·to_string()` was being parsed as a call to a path `tag·to_string`
+        // and dispatched to a built-in of the wrong arity. Nothing in the chain
+        // names a type, so it must stay a method call on `tag`.
+        let source = "\
+            scroll geom {
+                ☉ Σ P { ☉ x: f32 }
+                ⊢ P {
+                    ☉ rite new(x: f32) -> Self { Self { x } }
+                    ☉ rite scaled(self, k: f32) -> f32 { self.x * k }
+                }
+            }
+            rite main() {
+                ≔ p = geom·P·new(3.0);
+                ⤺ p·scaled(2.0);
+            }";
+        assert!(matches!(run(source), Ok(Value::Float(v)) if v == 6.0));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


paraphilic-ecchymosis:
Reviewed independently of the agent that wrote this — re-derived the numbers rather than confirming them.

**Lookahead indexing is right.** `check()` reads `self.current`; `peek_n(0)` is documented and implemented as the token *after* current. With current sitting on `·`, the scan starts on the following ident, as the comment claims.

**Test counts, both branches, same command (`RUST_MIN_STACK=33554432 cargo test --lib`, fresh worktrees):**

| | passed | failed | ignored |
|---|---|---|---|
| `develop` | 557 | 2 | 1 |
| this branch | 559 | 2 | 0 |

The 2 failures are `llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}` on both — pre-existing, untouched. Note for anyone reproducing: `stdlib::tests::test_parser_nested_brackets` blows the default thread stack on **both** branches, so bump `RUST_MIN_STACK` or you will misread a pre-existing limit as a regression here.

**The false-positive risk, measured.** The heuristic widens what `·` may consume, so the thing to prove is that no existing method call got captured. I diffed `sigil parse` output between the two binaries over **509 `.sigil`/`.sg` files** (nihil, qliphoth, morgoth, validation, examples, stdlib):

- **506 identical**
- **3 differ — all three fail to parse on `develop` and parse here**
- **0 regressions**

The three: `nihil/tomes/nihil-io/src/lib.sigil`, `nihil/tomes/nihil-models/tests/integration_test.sigil`, `qliphoth/packages/qliphoth-router/src/params.sigil`.

`params.sigil:152` is the whole change in one line:

```
≔ path! = format!("{}{}", location.pathname, new_query·to_string());
navigate(&path, crate·router·NavigateOptions {
```

`crate·router·NavigateOptions` becomes a path; `new_query·to_string()` beside it stays a method call. Both halves of the decision, same line, both right.

**Correcting my own issue #86:** I named three candidate commits and none of them was it. The culprit is `8f92d5c` (#61), found by bisect. I also called pre-#61 "working" — it was, by accident: the old guard was `consume_if(·) || (first_segment_is_type && consume_if(·))`, whose first operand always fires, so `first_segment_is_type` decided nothing and every `·` was eaten. #61 made the flag load-bearing for the first time and got the module case wrong. The branch-reconciliation handoff had already reached this conclusion — I had read it and did not connect it.

Approving and merging.

PR #89, opened 2026-09-10
fix(migrate): judge a closure parameter as a binding, not a field

### LINKED ISSUE #84: migrate: a closure parameter is classified as a field, so reserved names in closures go unwarned
Known gap in the reserved-name diagnostic added by #68. Recorded as an issue because the PR body is not where a live defect should live.

## The gap

The classifier decides a name's position from its **enclosing bracket**. For a closure parameter:

```rust
let f = |tome: u32| tome + 1;
```

the enclosing bracket is the *function body*, so `tome` is judged to be in **field** position rather than **binding** position.

That matters because those two positions disagree. `tome` — along with `ref`, `super`, `const`, `async`, `move`, `unsafe`, `volatile`, `atomic`, `simd` — is **accepted as a field name and refused as a binding**. So a reserved name used as a closure parameter is silently not warned about: a false negative, in the one direction the diagnostic exists to prevent.

## Why it was left

**Zero occurrences across the 6,091-file validation corpus** (crate registry plus this repository). So it is *unmeasured*, not known-harmless — there was no case available to test a fix against.

The obvious fix is a `|`-disambiguation heuristic in the bracket tracker. That was deliberately not written, because an untested heuristic in a scanner that already had six defects is a poor trade: #68's own history is that the scanner's bracket handling was the largest source of wrong answers, not the oracle.

## What a fix needs

1. A real corpus case, or a constructed test that is honest about being synthetic.
2. Bracket tracking that distinguishes `|params|` from bitwise-or and from a match guard's `|` alternation — the reason this is not a one-liner.
3. Verification that it does not regress the 0-false-positive result on the 6,091-file corpus, which is the property most worth protecting.

## Related

- #82 — the same false premise in the Python migrate tool, fixed in #83 (merged into #62's branch, not yet on `develop`).
- The position matrix this rests on: of 93 word-shaped lexer tokens, 58 are accepted everywhere, 15 refused everywhere, 10 refused as fields but fine as bindings, and 10 accepted as fields but refused as bindings. Lexing as a keyword is not the same as failing to parse.

paraphilic-ecchymosis:
Closes #84.

A closure parameter is a binding. The scanner was judging it as a struct field, because it took a name's position from the innermost unclosed bracket and a closure parameter list did not push one — so `|tome: u32|` written inside a function body was judged by that body's `{`.

## What the issue got right, and what it missed

The enclosing-bracket theory is exactly what the code does. `name_position` matches on `brackets.last()`, `(` is a parameter list and `{` is a struct body, and a closure pushed nothing of its own. Confirmed.

What the issue frames as a one-directional false negative is two-directional. Field and binding disagree **both ways** — that is the whole reason #68 carries a position at all:

| words | as a field | as a binding | a closure parameter was |
|---|---|---|---|
| `tome`, `ref`, `super`, `const`, `async`, `move`, `unsafe`, `volatile`, `atomic`, `simd` | accepted | refused | silently **not** warned |
| `self`, `Self`, `this`, `true`, `yay` (10 words) | refused | accepted | **wrongly** warned |

The second row is the half that actually occurs in real code. A corpus sweep found two of them:

- `cranelift-isle-0.113.1/src/trie_again.rs:469` — `let mut defer = |this: &Self, binding| {`
- `rustc-demangle-0.1.27/src/v0.rs:1133` — `let mut open_brace_if_outside_expr = |this: &mut Self| {`

Both were being reported as names the parser refuses. Both parse fine. So this was not only an unmeasured false negative — it was also a **live false positive**, in the one property #68 spent its effort on. Fixing the classification removes both.

## How the bar is disambiguated

`|` is three tokens. A `b'|'` frame is pushed onto the bracket stack only when the bar opens in **expression** position — after `=`, `(`, `[`, `{`, `,`, `;`, `=>`, `:`, another `|`, or the word `move` or `return`, or at the start of the file. A bitwise or and a pattern alternation both follow a *value* (identifier, literal, `)`, `]`), so neither is in that set.

The set is a deliberate allowlist, not "anything that is not a value". Misreading an or as a parameter list is the failure that costs something — a leaked frame would make every later `name:` in the file read as a parameter — so an unlisted predecessor is treated as an or, and the worst case is a closure that goes unjudged. `}` is left out for that reason.

Three things keep the stack balanced:

- `||` pushes nothing, whether it is a logical or or a zero-parameter closure. A closure declaring no parameters cannot produce a finding either way.
- A bar met while a `|` frame is on top **closes** it, restoring `in_where` the way every other close does.
- The frame is abandoned at the first `;`, `{`, `}`, `)`, `]` or `=>`, none of which a parameter list can contain. That is what makes a leading-bar match arm — `match x { | A => 0, | B => 1 }` — self-correcting rather than a stack leak.

The macro flag is inherited when the frame is pushed, so `foo!(|tome: u32| …)` stays as quiet as it was.

## Untyped parameters

Most closure parameters carry no type, so there is no colon for the scanner to find them by. Inside a `|` frame, and only there, a bare word is read as a parameter when it both **starts** one (the byte before it is the bar or a comma) and **ends** one (a comma or the bar follows).

Both halves are load-bearing. Without the first, the `u32` in `|x: u32|` is a word ending at a bar and gets reported as a parameter named after its own type; the `:` in front of it is what rules it out. `|x: &'a Tome|` fails the same test on the `a` of the lifetime. `|mut x|` is missed by it — a known, quiet false negative.

## Corpus

I could not find a checked-in corpus harness: nothing in the repo, in `tools/`, or in #68's commits runs one, and #68's 6,091-file figure is not reproducible from anything on the branch. So I reconstructed one — a temporary `#[test]` calling `collect_keyword_collisions` over every `.rs` file under `~/.cargo/registry/src` plus this repository, run on `origin/develop` and again on this branch, then removed before committing. It is a **superset** of #68's corpus, not the same set: the registry has grown since.

| | files scanned | files with findings | distinct name findings | total uses |
|---|---|---|---|---|
| `develop` @ 4b07854 | 21,897 | 42 | 44 | 125 |
| this branch | 21,897 | 40 | **42** | 123 |

**No finding was added.** The false-positive count stays at zero by inspection — every one of the 42 remaining is a genuine reserved name in a genuine declaring position (`simd` in `pulp`, `each` in `derive_builder`, `vary` in `tower-http`, `aspect` in `libc`'s `psp.rs` and in this repo, `alter` in `plurality/`). The two removed are the `this` false positives above.

## Tests

Seventeen new assertions across two tests in `migrate_span_tests`, covering all three meanings of the bar:

- `let f = |tome: u32| tome + 1;` warns — the issue's case
- `move |tome| …` warns, and `|a, tome: u32|` / `|a, tome|` warn with the name not directly after the bar
- a closure nested in another closure's body, both as `|a| |tome|` and as `|a| { let h = |tome| … }`
- `|x: u32|` does not report `u32`
- `let x = a | tome;` and `a | b | tome` report nothing, and a struct declared after them is still judged correctly — the stack-leak check
- `Some(tome) | None =>` is not misclassified, and a leading-bar match arm leaves the stack balanced
- `let f = || 1;` gives no panic and no finding, and a field judged afterwards proves the stack came back
- a bar inside a string literal and inside a comment is not code
- `|_: u32|` and `|_|` stay discards
- the two real `this` cases

**The `tome` half of this is synthetic and unmeasured.** #84 records zero occurrences across the 6,091-file corpus, and my 21,897-file re-sweep still finds none — no closure parameter named a word that binds worse than it fields. So the false-negative fix is untested against real code and remains so; what the corpus does establish is the other two properties: the `this` false positives are real and now gone, and the disambiguation adds nothing.

## Test counts

`RUST_MIN_STACK=33554432 cargo test --lib` in `parser/`: **561 passed, 2 failed** — identical to `develop` @ 4b07854, same two failures (`llvm_codegen::llvm::tests::test_generic_struct_basic`, `test_generic_struct_two_params`), unrelated to this change.

`cargo test --bin sigil`: **27 passed, 0 failed**, up from 25 on `develop`.

### DIFF
--- a/parser/src/main.rs
+++ b/parser/src/main.rs
@@ -5656,8 +5656,8 @@ fn parser_accepts(name: &str, position: NamePosition) -> bool {
 /// enclosing it.
 ///
 /// `None` means the context is not one of the three that can be judged — a
-/// closure parameter, a match arm, a type ascription somewhere unusual. Those
-/// are passed over rather than guessed at.
+/// match arm, a type ascription somewhere unusual. Those are passed over
+/// rather than guessed at.
 fn name_position(prev_words: [&str; 2], enclosing: Option<u8>) -> Option<NamePosition> {
     // A `where` bound is not a name being declared. `where Self: Sized` puts
     // `Self` in front of a colon inside a trait body, which otherwise reads as
@@ -5682,10 +5682,49 @@ fn name_position(prev_words: [&str; 2], enclosing: Option<u8>) -> Option<NamePos
         // A parameter list. `fn f(mut x: T)` lands here too, since `mut`
         // alone did not make it a local above.
         Some(b'(') => Some(NamePosition::Parameter),
+        // A closure parameter list, `|x: T|`. The scanner pushes a frame for
+        // it so that the enclosing bracket is the list rather than whatever
+        // block the closure was written inside — a function body's `{` read
+        // as a struct declaration, which judged a *binding* by the rules for
+        // a *field*. The two disagree for `tome`, `ref`, `super`, `const`,
+        // `async`, `move`, `unsafe`, `volatile`, `atomic` and `simd`: all ten
+        // are fine as a field and refused as a binding, so the name went
+        // unwarned. A closure parameter binds, so it is judged as one.
+        Some(b'|') => Some(NamePosition::Parameter),
         // A struct declaration or literal body.
         Some(b'{') => Some(NamePosition::Field),
         _ => None,
     }
+}
+
+/// Whether the `|` about to be read opens a closure parameter list, rather
+/// than being a bitwise or, or a pattern's alternation.
+///
+/// The three are told apart by what comes *before* the bar, because only one
+/// of them opens in expression position:
+///
+/// - `|x: T|` follows `=`, `(`, `[`, `{`, `,`, `;`, `=>`, `:`, another `|`,
+///   the word `move` or `return`, or the start of the file.
+/// - `a | b` follows a value — an identifier, a literal, `)` or `]`.
+/// - `Some(x) | None =>` follows a value-shaped token too.
+///
+/// The set is a deliberate allowlist rather than "anything that is not a
+/// value". Reading a bitwise or as a parameter list is the failure that costs
+/// something — it would leave a frame on the bracket stack — so an unlisted
+/// predecessor is treated as an or, and the worst case is a closure that goes
+/// unjudged. `}` is left out for the same reason: a statement can follow a
+/// block, but a closure written straight after one cannot.
+fn opens_closure_params(prev_byte: u8, prev_words: &[String; 2]) -> bool {
+    // `move` and `return` end in a letter, so the byte before the bar says
+    // nothing about them.
+    if prev_words[1] == "move" || prev_words[1] == "return" {
+        return true;
+    }
+    matches!(
+        prev_byte,
+        // `>` covers both `=>` and `->`; neither can precede a bitwise or.
+        0 | b'=' | b'(' | b'[' | b'{' | b',' | b';' | b'>' | b':' | b'|'
+    )
 }
 
 /// The scanner's state, which must survive from one code span to the next.
@@ -5711,6 +5750,14 @@ struct CollisionScan {
     /// what is really a macro argument. A struct literal nested inside the
     /// invocation pushes its own `{`, so it is still judged — only the
     /// innermost bracket decides.
+    ///
+    /// A closure's `|params|` pushes a `b'|'` frame, so that a closure
+    /// parameter is judged by the list it sits in rather than by the block
+    /// the closure was written inside. That frame is the only one whose
+    /// opening byte is ambiguous — `|` is also a bitwise or and a pattern's
+    /// alternation — so it is pushed only from expression position and
+    /// abandoned the moment something appears that a parameter list cannot
+    /// contain.
     brackets: Vec<(u8, bool, bool)>,
     /// The two words before the current one, for `let` and `let mut`.
     prev_words: [String; 2],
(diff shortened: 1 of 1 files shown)


paraphilic-ecchymosis:
Verified independently of the agent that wrote this, on a separate worktree and a separately drawn sample.

**Counts re-measured** (`RUST_MIN_STACK=33554432`, from `parser/`): lib **561 passed / 2 failed / 0 ignored**, bin **27 passed** (develop: 25). Same two pre-existing `llvm_codegen` generic-struct failures.

**The false positive is real, and I confirmed it on the actual file** rather than on the quoted snippet. `cranelift-isle-0.113.1/src/trie_again.rs:469` is verbatim `let mut defer = |this: &Self, binding| {`. Running the two binaries over it:

```
develop: warning: 1 name in this file is refused by the Sigil parser where it stands:
             this             1 use
PR89:    (silent)
```

`this` is refused as a field and accepted as a binding, so judging a closure parameter as a field reported valid code. That is a live defect in the 0-false-positive property #84 called the thing most worth protecting — and #84 did not mention it, because I wrote #84 counting only the direction that goes quiet.

**Independent regression sweep.** I drew my own 2,500-file sample from the registry and ran both binaries per file:

| | files with findings |
|---|---|
| develop | 4 |
| PR89 | 4 |
| **added by PR89** | **0** |
| removed by PR89 | 0 |

The four are `simd`+`this` in `pulp-0.22.2/src/x86/v3.rs`, `aspect` in `libc-0.2.189/src/psp.rs`, `vary` in `tower-http-0.6.7/src/cors/mod.rs`. I checked the `this` one that survives: `v3.rs:1197` is `this: V3,` — a genuine struct **field**, so the warning is correct and must stay. Good separation: the same word is now reported as a field and not as a closure parameter.

The two files this PR silences were not in my sample, which is what you would expect for 2 files out of ~21,897.

**Safety note for anyone reproducing this: `sigil migrate` rewrites files in place.** Do not point it at `~/.cargo/registry`. I swept on copies.

**Corpus-harness claim confirmed as a gap in my own earlier work.** There is no reusable harness behind #68's "6,091 files"; the number is not reproducible from anything on the branch. That is worth fixing separately — a diagnostic whose headline property is "0 false positives across a corpus" should ship the means to re-check it.

Approving and merging. The `|mut x|` gap is filed as a follow-up rather than left in a PR body.

PR #116, opened 2026-09-10
fix(interpreter): make run-dir deterministic — bind Self through a scroll, stop a HashMap picking the type

### LINKED ISSUE #104: `run-dir` is nondeterministic — same directory, same binary, alternating results
Reported by a session working on #95, reproduced against `aether`'s `aether-framework/aether-sigil/engine/math/src/{lib,vector}.sg`:

> Same directory, same binary: alternates between the right answer and `[R0000] Invalid struct operation`, **~40% failures**, identically on both branches.

It nearly produced a false A/B result during that work — a fix looked like it regressed something when the two runs simply landed differently.

## What I could NOT reproduce, and what I ruled out

I tried and failed to reproduce it with a minimal two-file case (a `lib.sigil` defining a struct with an impl, a `main.sigil` calling it): **12/12 runs identical**. So it needs something the minimal case lacks — more modules, generics, operator traits, or the specific shape of aether's math files.

Two plausible mechanisms I checked and **eliminated**, recorded so nobody re-checks them:

1. **File iteration order.** `run-dir` (`main.rs:669`) *does* sort, lib-first/main-last, and the comparator is consistent for the lib/main/other cases. Not the cause. (Note: `collect_sigil_files` at `main.rs:1013`, used by the workspace path, does **not** sort and pushes in raw `read_dir` order — that is a separate latent problem, not this one.)
2. **HashMap iteration order.** Rust's default hasher is randomly seeded per process, so iterating a `HashMap` during impl resolution would produce exactly this signature. But `ImplRegistry`'s `generic_impls` and `concrete_impls` are both `Vec`, so their order is stable. Not the cause.

So the mechanism is still unknown. Whoever picks this up should start from the aether math files rather than a synthetic case, since that is the only known reproducer.

## Why it matters more than it looks

`run-dir` is the entry point for every multi-module Sigil program, including `lares-server` and the aether Sigil port. A 40% failure rate that is invisible on a single run means:

- any "does this work?" check on a Sigil crate is a coin flip
- CI on Sigil code would flake without an obvious cause
- an A/B comparison between two builds is unsound unless each side is run repeatedly

That last one is the dangerous property: it silently corrupts exactly the kind of measurement used to validate compiler changes.

## Suggested first step

Run the aether math files under `run-dir` in a loop, capture both outcomes, and diff the interpreter state or `dump-ir` between a passing and a failing run. The error is `Invalid struct operation`, which suggests a type or impl lookup resolving differently — so something order- or identity-dependent is varying between runs even though the inputs do not.

paraphilic-ecchymosis:
Closes #104 (does not auto-close — this targets `develop`).

`run-dir` alternated between the right answer and `[R0000] Invalid struct
operation` on the same directory with the same binary, ~40% of runs. Two
defects compose; neither reproduces alone, which is why the minimal two-file
case in the issue came back 12/12 identical.

## 1. Trigger — `Self` is not bound through a module-qualified call

`type_name_for_self` (`eval_call`) read the type out of the **first** path
segment:

```rust
let first = &path.segments[0].ident.name;
if self.types.contains_key(first) { Some(first.clone()) } else { None }
```

Right for `P·new(…)`; wrong for `geom·P·new(…)`, where segment 0 is the scroll.
`Self` stayed unbound, so the `Self { .. }` literal in the callee fell back to
its raw name and the constructor returned a struct named, literally, `"Self"`.

`lookup_operator_impl` keys operator impls by the value's struct name, and
nothing is filed under `"Self"` — so the next `+` reported
`Invalid struct operation`. This half is deterministic.

## 2. Nondeterminism — recovering a `"Self"`-named value walked a `HashMap`

A `"Self"`-named value is recovered by matching its field names against every
registered struct. Both copies of that scan iterated `self.types` — a `HashMap`
— and returned on the first hit. The match is a *subset* test, so `{x}` matches
every struct with an `x`. With more than one candidate, the recovered type (and
so the name of whatever the method returns) changed per process.

That is the coin flip.

## Reproducer

```sigil
// shapes.sg — two structs of the same shape, one with an operator impl
☉ Σ P { ☉ x: f32 }
⊢ P {
    ☉ rite new(x: f32) -> Self { Self { x } }
    ☉ rite twice(self) -> Self { Self·new(self.x * 2.0) }
}
⊢ Add ∀ P {
    type Output = Self;
    rite add(self, o: Self) -> Self { Self·new(self.x + o.x) }
}
☉ Σ Q { ☉ x: f32 }
⊢ Q {
    ☉ rite new(x: f32) -> Self { Self { x } }
    ☉ rite twice(self) -> Self { Self·new(self.x * 2.0) }
}

// main.sg
☉ rite main() -> !i32 {
    ≔ a = shapes·P·new(1.0);   // (1) returns a struct named "Self"
    ≔ b = a·twice();           // (2) recovered as P, shapes·P, Q or shapes·Q
    ≔ c = b + b;               // succeeds only when (2) landed on P
    println(c·x);
    ⤺ 0;
}
```

Instrumented, the scan chose `P`, `shapes·P`, `Q` and `shapes·Q` across runs of
the same directory. Before: **8 passes in 20 runs**. After: **30 in 30**.

The same fix makes the reported aether case stable —
`vector·Vec3·new(…)` → `·normalize()` → `+` is now 12/12 identical and correct.

## The fix

- `type_name_for_self` takes the type from the path *prefix* (everything before
  the final segment), preferring its bare last segment — the name an
  unqualified call binds, and the name impl methods and operator impls are
  filed under — falling back to the fully qualified prefix.
- Both recovery scans go through a new `struct_types_matching_fields`, which
  returns candidates in a stable order: exact field-set matches first, then
  supersets, each group sorted by name.
- The enum-variant suffix scan in `eval_path` had the same shape (first hit
  while walking `self.types`) and is collected and sorted the same way.

Fixing only the trigger would leave the coin flip in place for any other route
to a `"Self"`-named value; fixing only the scan would make the failure
deterministic rather than absent.

This is entirely in name resolution — no parser guard was touched. Consistent
with `docs/findings/middledot-module-path-vs-method-call.md`, which is why the
`geom·P·new` path already parses as a path (its rule 4).

## Ruled out (recorded so it is not re-checked)

- **File iteration order.** `run_directory` sorts lib-first / main-last,
  consistently for those cases.
- **Impl-registry order.** `generic_impls`, `concrete_impls` and the
  `operator_impls` candidate lists are all `Vec`s.

## Tests

| | develop | this branch |
|---|---|---|
| `RUST_MIN_STACK=33554432 cargo test --release --no-fail-fast` (lib) | 569 passed / 2 failed | **571 passed / 2 failed** |
| bin | 41 passed / 0 failed | 41 passed / 0 failed |
| `jormungandr/tests/run_tests_rust.sh` | 797 passed / 4 failed / 2 skipped | 797 passed / 4 failed / 2 skipped |

Two regression tests added. `an_associated_function_reached_through_its_scroll_binds_self`
was confirmed to fail on unmodified `develop` before being kept.

`docs/findings/run-dir-nondeterminism-self-named-values.md` records the
mechanism, the reproducer and the general rule: a `HashMap` may not decide name
resolution.

### DIFF
--- a/docs/findings/run-dir-nondeterminism-self-named-values.md
+++ b/docs/findings/run-dir-nondeterminism-self-named-values.md
@@ -1,0 +1,117 @@
+# Why `run-dir` alternated between the right answer and "Invalid struct operation"
+
+**Recorded:** 2026-09-10, closing #104.
+
+Same directory, same binary, ~40% of runs failing with
+`[R0000] Invalid struct operation`. Reported against `aether`'s
+`engine/math/src/{lib,vector}.sg`. It nearly produced a false A/B result: a fix
+looked like a regression when the two runs simply landed differently.
+
+Two defects compose. Neither is sufficient alone, which is why a minimal
+two-file case did not reproduce it and the real one did.
+
+## 1. The trigger: `Self` is not bound through a module-qualified call
+
+`eval_call` decides what `Self` means for the callee from the call path
+(`interpreter.rs`, `type_name_for_self`). It tested only the **first** segment:
+
+```rust
+let first = &path.segments[0].ident.name;
+if self.types.contains_key(first) { Some(first.clone()) } else { None }
+```
+
+That is right for `P·new(…)` and wrong for `geom·P·new(…)`, where segment 0 is
+the scroll. `Self` stayed unbound, and a `Self { .. }` literal in the callee
+falls back to its own raw name — so the constructor returned a struct named,
+literally, **`"Self"`**.
+
+Nothing is filed under `"Self"`. `lookup_operator_impl` keys operator impls by
+the value's struct name, so the next `+` on that value found no `impl Add` and
+reported `Invalid struct operation`.
+
+This half is deterministic: with two modules it fails every time.
+
+## 2. The nondeterminism: recovering a `Self`-named value walked a `HashMap`
+
+A `Self`-named value is not a dead end — a method call on one is recovered by
+matching its field names against every registered struct and using the first
+type that fits. That scan (two copies, both in `eval_method_call`) iterated
+`self.types`, a `HashMap`, and returned on the first hit:
+
+```rust
+for (type_name, type_def) in &self.types {
+    …
+    let matches = field_names.iter().all(|f| def_fields.contains(f));
+    if matches { /* call type_name·method, bind Self to type_name */ }
+}
+```
+
+`HashMap` iteration order is seeded per process. The match is also a **subset**
+test, so `{x}` matches every struct that has an `x`. With more than one
+candidate the recovered type — and therefore the name of whatever the method
+returns — changed from run to run.
+
+That is the coin flip. In the reduced case below the scan lands on `P`,
+`shapes·P`, `Q` or `shapes·Q`; only bare `P` carries the `Add` impl, so the
+program worked in 8 runs out of 20 and reported `Invalid struct operation` in
+the other 12.
+
+```sigil
+// shapes.sg — two structs of the same shape, one with an operator impl
+☉ Σ P { ☉ x: f32 }
+⊢ P {
+    ☉ rite new(x: f32) -> Self { Self { x } }
+    ☉ rite twice(self) -> Self { Self·new(self.x * 2.0) }
+}
+⊢ Add ∀ P {
+    type Output = Self;
+    rite add(self, o: Self) -> Self { Self·new(self.x + o.x) }
+}
+☉ Σ Q { ☉ x: f32 }
+⊢ Q {
+    ☉ rite new(x: f32) -> Self { Self { x } }
+    ☉ rite twice(self) -> Self { Self·new(self.x * 2.0) }
+}
+
+// main.sg
+☉ rite main() -> !i32 {
+    ≔ a = shapes·P·new(1.0);   // (1) returns a struct named "Self"
+    ≔ b = a·twice();           // (2) recovered as P, shapes·P, Q or shapes·Q
+    ≔ c = b + b;               // succeeds only when (2) landed on P
+    println(c·x);
+    ⤺ 0;
+}
+```
+
+## What was ruled out
+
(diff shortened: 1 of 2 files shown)


paraphilic-ecchymosis:
Verified independently, and the reproduction is stronger than the issue reported.

I built the branch and ran your scroll-qualified case 25 times against each binary:

| | 25 runs |
|---|---|
| `develop` | **0 pass / 25 fail** — `Invalid struct operation` |
| this branch | **25 pass / 0 fail** — returns `5` |

Note it was 25/25 failing rather than the ~40% in #104: my fixture declares three same-shape structs in the scroll, so the `HashMap` subset-scan almost always lands on the wrong one. That matches your analysis exactly — the flake rate is a function of how many candidate structs share the field shape, which is why the minimal two-file case in the issue came back 12/12 clean. Good root-cause work; the composition of a deterministic trigger with a nondeterministic recovery is why neither half reproduced alone.

Lib suite on this branch: **571 passed, 2 failed, 0 ignored** — the two being the pre-existing `llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}`. Baseline before this was 569/2, so +2 are yours. CI green on all four checks.

**One correction I owe you.** Issue #104 told you HashMap iteration order was *ruled out*:

> Rust's default hasher is randomly seeded per process, so iterating a `HashMap` during impl resolution would produce exactly this signature. But `ImplRegistry`'s `generic_impls` and `concrete_impls` are both `Vec`, so their order is stable. Not the cause.

The mechanism was right and I checked the wrong container. The offending map is `self.types` in the interpreter, not the impl registry. That note could have steered you away from the actual cause — you found it anyway, but the issue was actively misleading and I have corrected it there.

Worth recording somewhere durable: this bug was *latent* until module-qualified calls started working. #87 and #103 made `geom·P·new(…)` resolve, which is precisely the shape that leaves `Self` unbound. Fixing a parser gap exposed an interpreter one that had never been reachable before.

Approving and merging.

PR #119, opened 2026-09-10
fix(cli): run-dir loads a directory's modules without running its entry point

### LINKED ISSUE #107: run-dir: main() executes once per module plus one (N files → N+1 runs; single file runs twice)
## Summary

`sigil run-dir <dir>` executes `main()` **once per module in the directory,
plus one** — `N` files produce `N+1` invocations. It should run exactly once.

A single-file directory runs `main()` **twice**.

## Reproduction

`app.sg` in an otherwise empty directory:

```sigil
rite main() {
    println("hi");
}
```

```console
$ sigil run-dir .
hi
hi
```

Add inert modules (`rite unusedN() { ↩ N; }`), none of which define `main`:

| `.sg` files in dir | times `hi` printed | expected |
|---|---|---|
| 1 | 2 | 1 |
| 2 | 3 | 1 |
| 3 | 4 | 1 |
| 4 | 5 | 1 |
| 5 | 6 | 1 |

The extra modules are never imported and define nothing that is used. The count
tracks file count alone.

Deterministic — 5/5 identical runs at each size.

## Regression

| build | 1 file | 2 files |
|---|---|---|
| `develop` @ `c6ef995` | 2 | 3 |
| 2026-03-02 build | 1 | 2 |

Two distinct defects are visible here:

1. **Long-standing:** `main()` is run once per loaded module rather than once
   per program. Present in the March build (2 files → 2 runs).
2. **Newer:** current `develop` adds one further invocation across the board
   (`N` → `N+1`), so even the single-file case is now wrong.

## Impact

This makes `run-dir` unusable as a test entry point for a multi-module project,
which is precisely the case it exists for. Real example — morgoth's Phase 28
integration test with its 17 sibling modules present emits its 9-line expected
output 19 times (171 lines), so any output comparison fails.

It also means any `main()` with side effects (writing a file, opening a socket,
spawning a process) performs them `N+1` times. Morgoth's `main()` spawns PTYs
and child processes, so `run-dir src` on the real tree is not merely noisy.

Related to but distinct from #104 (`run-dir` nondeterminism): this one is
perfectly reproducible.

---

paraphilic-ecchymosis:
Closes #107 (does not auto-close — this targets `develop`).

### Rebase note

Rebased twice: first onto `f9bafb6` (was 14 behind), then onto `aa3d039`
(9 behind) as `develop` kept moving. The **second rebase applied cleanly** — no
conflict. The first had one, in `parser/src/main.rs`,
and it was mine on both sides: this branch appends `mod run_directory_tests` at
the end of the file, and #133/#135 landed `mod collision_corpus_tests` in the
same place. Two independent `#[cfg(test)]` modules, neither replacing the
other — **both kept**, nothing else in the conflict region. The functional
hunks (`execute_definitions`, the hoisted `load_rank`) applied cleanly.

Re-verified after the rebase rather than assuming: the N-file table below still
holds, and #128's nested-module resolution still works on top of it (a
`tui/{mod,grid,vterm,layout}.sg` + `tui/widgets/button.sg` tree runs correctly).

## Cause

`Interpreter::execute` (`interpreter.rs:3110`) does not only execute the file it
is handed. After running the file's items it looks `main` up **in the globals**
and calls it:

```rust
for item in &file.items {
    result = self.execute_item(&item.node)?;
}

let main_fn = self.globals.borrow().get("main")…;
if let Some(f) = main_fn {
    if f.params.is_empty() {
        result = self.call_function(&f, vec![])?;
    }
}
```

Because it looks in the globals rather than in the file, once *any* file has
registered `main`, every later `execute` runs it again. `run_directory` called
`execute` once per file and then `call_main` — hence N + 1.

`execute_definitions` (`interpreter.rs:3138`) is exactly `execute` minus the
auto-call, and its doc comment already says "Use this when loading files as part
of a multi-file workspace". The workspace loader (`main.rs:978`, `main.rs:1148`),
the module loader and the stdlib loader all use it. `run_directory` was the one
caller left on `execute`.

## Second defect in the same function

The load-order comparator matched name pairs:

```rust
match (a_name, b_name) {
    ("lib.sigil", _) | ("lib.sg", _) => Ordering::Less,
    (_, "lib.sigil") | (_, "lib.sg") => Ordering::Greater,
    ("main.sigil", _) | ("main.sg", _) => Ordering::Greater,
    (_, "main.sigil") | (_, "main.sg") => Ordering::Less,
    _ => a_name.cmp(b_name),
}
```

`cmp("lib.sg", "lib.sigil")` and `cmp("lib.sigil", "lib.sg")` both hit arm 1 and
both return `Less`; both directions of the `main.*` pair hit arm 3 and both
return `Greater`. That is not a total order, and `slice::sort_by` is documented
as permitted to panic on one. Replaced with a rank — interface 0, modules 1,
entry point 2 — and the file name as tie-break.

This is *not* what made #104 flake; the ordering it produced was still
deterministic.

## Measured

`app.sg` plus N−1 inert modules that define nothing used, counting how often
`main`'s `println` fires:

| files in dir | before | after | expected |
|---|---|---|---|
| 1 | 2 | **1** | 1 |
| 2 | 3 | **1** | 1 |
| 3 | 4 | **1** | 1 |
| 4 | 5 | **1** | 1 |
| 5 | 6 | **1** | 1 |

A directory holding both `main.sg` and `main.sigil` ran `main` three times
before; once now.

## Tests

Rebased again onto `develop` @ `aa3d039`. Re-measured on both sides:

| | develop @ aa3d039 | this branch |
|---|---|---|
| `RUST_MIN_STACK=33554432 cargo test --release --no-fail-fast` (lib) | 586 passed / 2 failed | 586 passed / 2 failed |
| bin | 46 passed / 0 failed | **49 passed / 0 failed** |
| `jormungandr/tests/run_tests_rust.sh` | 801 / 4 failed / 2 skipped | 801 / 4 failed / 2 skipped |
| morgoth `run_tests.sh` @ `adad4cf` | — | 239 / 0 |

Three tests added. `loading_a_module_does_not_run_the_entry_point` pins both
halves of the contract `run_directory` now depends on — that `execute` *does*
invoke `main` (so a future change there does not silently reintroduce this) and
that `execute_definitions` does not, neither on the entry file nor on a sibling
loaded after it. The two ordering tests document the total-order property rather
than reproducing the old comparator, which was inline and is gone.

Independent of #116 (#104): different file, no overlap.

### DIFF
--- a/parser/src/main.rs
+++ b/parser/src/main.rs
@@ -647,6 +647,16 @@ fn run_file(path: &str, program_args: &[String], skip_typecheck: bool) -> ExitCo
     }
 }
 
+/// Where a file sits in a directory's load order: the tome's interface, then
+/// its other modules, then the entry point.
+fn load_rank(name: &str) -> u8 {
+    match name {
+        "lib.sigil" | "lib.sg" => 0,
+        "main.sigil" | "main.sg" => 2,
+        _ => 1,
+    }
+}
+
 /// Run all .sg/.sigil files in a directory as a multi-module program
 fn run_directory(dir_path: &str, program_args: &[String]) -> ExitCode {
     use std::path::Path;
@@ -670,16 +680,17 @@ fn run_directory(dir_path: &str, program_args: &[String]) -> ExitCode {
     // 1. lib.sigil first (defines the module's public interface)
     // 2. Other modules in alphabetical order
     // 3. main.sigil last (uses definitions from other modules)
+    //
+    // Ranked rather than matched pairwise. The pairwise form answered `Less`
+    // to both `cmp("lib.sg", "lib.sigil")` and `cmp("lib.sigil", "lib.sg")`,
+    // and `Greater` to both directions of the `main.*` pair, which is not a
+    // total order -- `sort_by` is allowed to panic on one.
     files.sort_by(|a, b| {
         let a_name = Path::new(a).file_name().and_then(|n| n.to_str()).unwrap_or("");
         let b_name = Path::new(b).file_name().and_then(|n| n.to_str()).unwrap_or("");
-        match (a_name, b_name) {
-            ("lib.sigil", _) | ("lib.sg", _) => std::cmp::Ordering::Less,
-            (_, "lib.sigil") | (_, "lib.sg") => std::cmp::Ordering::Greater,
-            ("main.sigil", _) | ("main.sg", _) => std::cmp::Ordering::Greater,
-            (_, "main.sigil") | (_, "main.sg") => std::cmp::Ordering::Less,
-            _ => a_name.cmp(b_name),
-        }
+        load_rank(a_name)
+            .cmp(&load_rank(b_name))
+            .then_with(|| a_name.cmp(b_name))
     });
 
     if files.is_empty() {
@@ -762,8 +773,13 @@ fn run_directory(dir_path: &str, program_args: &[String]) -> ExitCode {
             }
         };
 
-        // Execute to register all definitions
-        if let Err(e) = interpreter.execute(&ast) {
+        // Register this file's definitions *without* running the entry point.
+        // `execute` looks `main` up in the globals rather than in the file it
+        // was handed, so once any file had defined it, every later file's load
+        // ran it again -- and `call_main` below then ran it once more. A
+        // directory of N modules invoked `main` N+1 times, repeating every side
+        // effect it performed.
+        if let Err(e) = interpreter.execute_definitions(&ast) {
             eprintln!("Error loading '{}': {}", file_path, e);
             return ExitCode::from(1);
         }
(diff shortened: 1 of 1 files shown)


paraphilic-ecchymosis:
Verified with an observable side effect rather than by reading the code path:

```sigil
// src/main.sg
rite main() { println("MAIN RAN"); ⤺ 0; }
```

| | `MAIN RAN` printed |
|---|---|
| `develop` | **2 times** |
| this branch | **1 time** |

Lib suite 569 passed / 2 failed, the two pre-existing `test_generic_struct_*`.

**Every `run-dir` invocation has been running the program twice**, doubling every side effect it performs. `Interpreter::execute` does not only execute the file it is handed — after running the items it looks `main` up in the globals and calls it — so `run-dir`, which calls `execute` per file to register definitions and then calls `main` deliberately, got two runs.

This is worse than it sounds for anything that writes. A program that appends to a file, increments a counter, sends a request or creates a resource did all of it twice, silently, with a correct-looking exit code. It has been invisible because the most common shapes hide it: a pure computation returns the same value both times, and a blocking server never returns from the first call so the second never happens. `lares-server` is exactly that second case — `PORT=7301 sigil run-dir src` looked fine only because `main` blocks in its accept loop.

Same family as #131: `run-dir` reported "ran your program" while doing something narrower and different, and the report was in the reader's vocabulary.

Approving and merging.

Issue #112, opened 2026-09-10
migrate: JS numeric literals emit as `I64` beside `f64`-typed props — `pct >= 70` fails to check


paraphilic-ecchymosis:
Found re-running the Lares client regeneration against the current `lares` UI, with a compiler from `develop` @ `c6ef995` (`--features minimal,protocols,react-migrate`).

`sigil migrate --from-react` emits JavaScript numeric literals as **integer** Sigil literals even when they sit next to a value it has itself typed as `f64`. Sigil has no int→float literal adaptation, so the generated file does not check.

## Reproduction

React (`lares/ui/src/tabs/rag-tab.tsx:147`):

```tsx
function ScoreBar({ score }: { score: number }) {
  const pct = Math.max(0, Math.min(100, score * 100))
  const color = pct >= 70 ? "var(--success)" : pct >= 40 ? "var(--warning-text)" : "var(--error)"
```

Generated:

```sigil
☉ rite score_bar(score: f64) -> VNode! {
    ≔ pct = math_max(0, math_min(100, score * 100));
    ≔ color = ⎇ pct >= 70 { "var(--success)" } ⎉ { ⎇ pct >= 40 { … } };
```

```
$ sigil check output/score-bar.sigil
[E0003] Error: comparison operands must have same type: left=Float(F64), right=Int(I64)
error: aborting due to 2 previous errors
```

`score * 100` makes `pct` an `F64`; `70` and `40` are `I64`.

## The underlying rule, isolated

```sigil
rite main() {
    ≔ x: f64 = 0.5;
    ⎇ x ≥ 70 { println("hi"); }        // E0003, exit 1
}
```

```sigil
rite main() {
    ≔ x: f64 = 0.5;
    ⎇ x ≥ 70.0 { println("hi"); } ⎉ { println("lo"); }   // clean, exit 0
}
```

So `70.0` is all that is wanted. This is the migrator's job rather than the checker's: Sigil's strictness here looks deliberate (it is the same strictness recorded as S124), and JS has a single `number` type, so the translation from one to the other is where the float-ness has to be decided.

`math_max`/`math_min` themselves resolve fine — verified separately, so they are not implicated.

## Why this appeared only now

It is a consequence of the prop-type extraction fix. While every prop was `Any`, `score` was `Any`, `score * 100` was `Any`, and nothing collided — the counterfactual checks clean:

```sigil
☉ rite score_bar(score: Any) -> i64! {
[...]

PR #127, opened 2026-09-10
fix(typeck): resolve Self to the implementing type for primitive impls

### LINKED ISSUE #108: `Self` does not resolve inside `⊢ Trait ∀ <primitive>` — yields "expected bool!, found bool!"
Found while re-measuring the Lares/Qliphoth capability probe against `develop` @ `c6ef995`, compiler built `--no-default-features --features minimal,protocols`.

`Self` in the method body of `⊢ Trait ∀ <type>` resolves correctly when the impl target is a **named** type, and does **not** resolve when the target is a **primitive**. The result is a type error where the two sides are the same type — sometimes printed identically, which makes the message read as nonsense.

## Reproduction

```sigil
Θ IntoJs {
    rite into(self: Self!) -> i64!;
}

rite take_bool(b: bool!) -> i64! { 0 }
rite take_f64(n: f64!) -> i64! { 0 }

⊢ IntoJs ∀ bool {
    rite into(self: Self!) -> i64! {
        take_bool(self)
    }
}

⊢ IntoJs ∀ f64 {
    rite into(self: Self!) -> i64! {
        take_f64(self)
    }
}

rite main() {
    println("ok");
}
```

```
$ sigil check self-primitive.sigil
[E0003] Error: type mismatch in argument 1: expected bool!, found bool!
[E0003] Error: type mismatch in argument 1: expected F64!, found f64!
error: aborting due to 2 previous errors
```

Exit 1 on 5/5 runs (checked explicitly — #104's nondeterminism does not apply to `check`).

## The bisect

Two one-line variants, both **clean** (exit 0 on 5/5):

**Write the primitive instead of `Self`** — so the impl target is fine, and only `Self` is at fault:

```sigil
⊢ IntoJs ∀ bool {
    rite into(self: bool!) -> i64! { take_bool(self) }   // clean
}
```

**Keep `Self`, impl for a struct** — so `Self` is fine, and only the primitive target is at fault:

```sigil
Σ Wrap { x: i64! }
⊢ IntoJs ∀ Wrap {
    rite into(self: Self!) -> i64! { take_wrap(self) }   // clean
}
```

So it is specifically **`Self` × primitive target**.

## What the messages are telling you

`expected F64!, found f64!` is the readable one: the *declared parameter* normalises to the primitive `F64`, while `Self` comes back as an unresolved `Named("f64")`. They never unify.

`expected bool!, found bool!` is the **same defect**, and the identical rendering is a second-order symptom — `Bool` and `Named("bool")` both print as `bool`. Anyone who meets that message first will reasonably assume a printer bug and look in the wrong place; it is the resolver.

Same shape as S22 (lifetimes not unifying with themselves), where the message also pointed at the wrong half of the type.

## Where it bites

`qliphoth/packages/qliphoth-sys/src/js.sigil` on qliphoth `develop` @ `2ab768b` — the JS-interop boundary. Six `⊢ Into[JsValue] ∀ …` impls over `String`, `&str`, `bool`, `f64`, `i32`, `u32`; five of them error, and all five are this:

```
expected &str!, found Unknown!      (∀ &str — the String case resolves to Unknown rather than Named)
expected bool!, found bool!
expected F64!, found f64!
expected I32!, found i32!
expected U32!, found u32!
```

This is the only file in the 52-file Qliphoth framework set whose failure is a **compiler** defect rather than unported Rust, and it is the file that converts Sigil primitives into JS values — so it is on the path of any Qliphoth app that passes a primitive to the host.

The `∀ String` → `Unknown!` variant suggests the resolver has at least two failure modes here (unresolved-named vs. no type at all), which may or may not be one fix.

## Note on reading the output

The spans on all of these point at the closing brace *after* the failing call, not at the call. That is filed separately — it is not specific to this defect, and it made this one considerably harder to locate.

paraphilic-ecchymosis:
Closes #108. (`Closes` does not auto-close in this repo — I will close it by hand once this merges.)

## The defect

`⊢ Into[JsValue] ∀ f64` bound `Self` to `Named("f64")`. There is no unify arm between that and `Type::Float(F64)`, so every use of `self` in the body failed against its own type:

```
type mismatch in argument 1: expected F64!, found f64!
```

For `bool` the two render identically, which is how the original report read — `expected bool!, found bool!`. A type mismatched against itself sends the reader after a printer bug; it is the resolver. Same shape as S22, where the message also pointed at the wrong half of the type.

A reference target failed differently and worse. `∀ &str` is not a `TypeExpr::Path`, so `type_path_to_name` fell to its catch-all and returned the literal string `"Unknown"`:

```
type mismatch in argument 1: expected &str!, found Unknown!
```

## The fix

Both `current_self_type` assignment sites built `Type::Named { name, generics }` unconditionally from that string. `convert_type` already maps primitives, references, tuples and the rest correctly — it is what every other type position goes through — so the impl target now goes through it too:

```rust
self.current_self_type = Some(match self.convert_type(&impl_block.self_ty) {
    Type::Named { .. } => Type::Named { name: type_name, generics: generic_types },
    primitive_or_ref => primitive_or_ref,
});
```

Only user-defined types come back as `Named`, and **that arm is left byte-identical** — so nothing about generic or struct impls changes. That was deliberate: routing everything through `convert_type` would also have changed the bare-`Foo`-with-generic-params case, which `unify` currently accepts through its empty-generics rule, and this issue is not the place to disturb that.

Bisected in both directions before fixing: writing `bool!` in place of `Self!` is clean, and `Self!` in an impl for a struct is clean, so it is specifically `Self` × a non-`Named` target.

## Why it is worth the churn

It was the **only compiler-caused failure in Qliphoth's entire framework set** — the other twelve are unported Rust ([qliphoth#17](https://github.com/Daemoniorum-LLC/qliphoth/issues/17)). And it sits on `packages/qliphoth-sys/src/js.sigil`, the Sigil-primitive-to-JS-value boundary that every Qliphoth app passing a primitive to the host goes through.

## Validation

Each side measured, not assumed:

| | |
|---|---|
| minimal repro | exit 1 → exit 0, 3/3 runs each side |
| controls | `bool!` for `Self!`, and `Self!` over a struct — clean before and after |
| qliphoth `develop` @ `2ab768b` | **39/52 → 40/52** framework files; exactly one file changed status, `js.sigil`, and it is the intended one |
| lares-client `output/` | 101/102 before and after, **identical failure set** (the one failure is #112, unrelated) |
| `cargo test --lib` | **524 passed, 0 failed** |
| Sigil suite | **798 passed, 4 failed** |

The four suite failures are `P0_005_kafka_producer`, `P0_007_amqp`, `P1_006_kafka_consumer`, `P1_008_amqp_consumer` — all `Runtime error` from absent brokers, all environmental, none a type error.

Unit-test and suite counts are under `--features minimal,protocols`, since the default features need LLVM 18; the 885 quoted in the Lares handoff is the full feature set. Worth a CI confirmation under `jit,llvm,wasm,lsp,protocols` rather than taking my numbers as covering it.

## Regression test

`jormungandr/tests/spec/25_nihil_gaps/test_self_in_primitive_impl.sg` pins all five original error shapes — `bool`, `f64`, `i32`, `u32`, `&str` — and carries a named-type impl as a control for the arm this change deliberately does not touch. It fails to type-check on the parent commit with exactly the five errors from the issue.

## Coordination

Found by the Lares capability probe (gap **S125**). `sigil-lang` had another active session on it (`lr-adhoc-7`, `sigil-tui-support`), so I checked for overlap before starting — that worktree was at `4461b5d` with no uncommitted changes and nothing ahead of `origin/develop`, and had claimed a worktree path rather than files. Claimed `parser/src/typeck.rs` on the Lares coord-log before touching it. This work was done in a separate detached worktree off `origin/develop`; the shared `sigil-lang` checkout (on `feature/s1-sigil-web-wasm-compat`, dirty) was not touched.

### DIFF
--- a/jormungandr/tests/spec/25_nihil_gaps/test_self_in_primitive_impl.sg
+++ b/jormungandr/tests/spec/25_nihil_gaps/test_self_in_primitive_impl.sg
@@ -1,0 +1,97 @@
+// Test: `Self` inside a trait impl for a primitive or reference type
+// Issue: sigil-lang#108 (Lares capability probe, gap S125)
+// Priority: P1 - blocks the Sigil-primitive → JS-value boundary in Qliphoth
+//
+// Purpose:
+// `⊢ Trait ∀ f64` must bind `Self` to the primitive `f64`. It used to bind it
+// to `Named("f64")`, which has no unify arm against `Type::Float(F64)`, so
+// every use of `self` in the body failed against its own type:
+//
+//     type mismatch in argument 1: expected F64!, found f64!
+//
+// For `bool` the two render identically, which is why the original report read
+// as nonsense — "expected bool!, found bool!". A reference target was worse:
+// `∀ &str` is not a `TypeExpr::Path`, so the impl type resolved to the literal
+// string "Unknown".
+//
+// The regression this pins is a *type* check, so the assertions below are
+// incidental — the test fails at check time if the gap reopens.
+//
+// Expected behavior:
+// All assertions pass; the file type-checks.
+
+Θ Describe {
+    rite describe(self: Self!) -> i64!;
+}
+
+rite take_bool(b: bool!) -> i64! { ⎇ b { 1 } ⎉ { 0 } }
+rite take_f64(n: f64!) -> i64! { ⎇ n > 0.0 { 1 } ⎉ { 0 } }
+rite take_i32(n: i32!) -> i64! { ⎇ n > 0 { 1 } ⎉ { 0 } }
+rite take_u32(n: u32!) -> i64! { ⎇ n > 0 { 1 } ⎉ { 0 } }
+rite take_str(s: &str!) -> i64! { ⎇ s == "" { 0 } ⎉ { 1 } }
+
+⊢ Describe ∀ bool {
+    rite describe(self: Self!) -> i64! { take_bool(self) }
+}
+
+⊢ Describe ∀ f64 {
+    rite describe(self: Self!) -> i64! { take_f64(self) }
+}
+
+⊢ Describe ∀ i32 {
+    rite describe(self: Self!) -> i64! { take_i32(self) }
+}
+
+⊢ Describe ∀ u32 {
+    rite describe(self: Self!) -> i64! { take_u32(self) }
+}
+
+⊢ Describe ∀ &str {
+    rite describe(self: Self!) -> i64! { take_str(self) }
+}
+
+// A named type must keep resolving exactly as before — this is the control,
+// and it is the path the fix deliberately leaves untouched.
+Σ Wrap { x: i64! }
+rite take_wrap(w: Wrap!) -> i64! { 1 }
+
+⊢ Describe ∀ Wrap {
+    rite describe(self: Self!) -> i64! { take_wrap(self) }
+}
+
+rite test_primitive_impls() {
+    ≔ b: bool = true;
+    assert_eq(take_bool(b), 1);
+
+    ≔ f: f64 = 2.5;
+    assert_eq(take_f64(f), 1);
+
+    ≔ i: i32 = 7;
+    assert_eq(take_i32(i), 1);
+
+    ≔ u: u32 = 7;
+    assert_eq(take_u32(u), 1);
+
+    println("  PASS: test_primitive_impls");
+}
+
+rite test_reference_impl() {
+    assert_eq(take_str("hello"), 1);
+    assert_eq(take_str(""), 0);
+    println("  PASS: test_reference_impl");
+}
+
+rite test_named_impl_unchanged() {
+    ≔ w = Wrap { x: 1 };
+    assert_eq(take_wrap(w), 1);
+    println("  PASS: test_named_impl_unchanged");
+}
+
+☉ rite main() -> !i32 {
+    println("=== S125: Self inside a primitive trait impl ===");
+    test_primitive_impls();
+    test_reference_impl();
+    test_named_impl_unchanged();
+    println("All Self-in-primitive-impl tests passed!");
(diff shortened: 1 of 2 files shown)


paraphilic-ecchymosis:
Verified with your own reproduction, after failing to construct one myself — which is worth recording, because it says something about the defect.

```
develop  exit=1   [E0003] type mismatch in argument 1: expected bool…
PR127    exit=0   ✓ no errors
```

Lib suite 569 passed / 2 failed, the two being the pre-existing `llvm_codegen::llvm::tests::test_generic_struct_*`.

I first tried three constructions from the PR description — `⊢ Into[JsValue] ∀ f64` using `self` in the body, then passing `self` as an argument, then a `∀ &str` target — and **all three passed on both builds**. Only your committed spec file reproduced it. That is not a criticism of the write-up; it means the trigger is narrower than the prose implies, and it is an argument for the test file being the artefact that matters rather than the description. Anyone re-deriving this from the issue text alone would likely conclude it was already fixed.

**The diagnostic angle is the part I would keep visible.** `expected bool!, found bool!` is a message that sends the reader after a printer bug — the two halves render identically because one is `Named("bool")` and the other is `Type::Bool`, and the renderer has no way to say so. The `∀ &str` case is worse in a more useful way: `type_path_to_name` falls to its catch-all and the impl type resolves to the literal string `"Unknown"`, which at least *looks* wrong rather than looking like nonsense.

That is the same shape as S22 and, more broadly, as #131 — a report that answers a narrower question than the reader believes it is answering. Here the message answers "do these two types unify" and renders as though it were answering "are these two types the same", and the reader cannot tell the difference.

Approving and merging.

PR #128, opened 2026-09-10
feat(interpreter): reach a module in a subdirectory of its tome

### LINKED ISSUE #109: Modules must be flat: subdirectories inside a tome are unreachable (blocks grouping a reusable TUI layer)
## Summary

A tome is a **flat** directory of `.sg` files. Modules in a subdirectory are
unreachable: neither a nested path (`invoke tome·tui·greet·{greet}`) nor a flat
one (`invoke tome·greet·{greet}`) resolves them, and `run-dir` does not recurse
into subdirectories at all.

This puts a ceiling on how large a single Sigil program can grow before its
module list becomes unnavigable, since grouping is impossible.

## Reproduction

```
subdir/
├── app.sg
└── tui/
    └── greet.sg
```

`tui/greet.sg`
```sigil
rite greet() {
    ↩ "from-subdir";
}
```

`app.sg`
```sigil
invoke tome·tui·greet·{greet};

rite main() {
    println(greet());
}
```

```console
$ cd subdir && sigil run ./app.sg
Runtime error in './app.sg': [R0003] Runtime error: undefined variable: `greet`
```

Same with the flat spelling `invoke tome·greet·{greet};` — also unresolved.

`run-dir` does not see the file either:

```console
$ sigil run-dir subdir
Loading 1 modules from '.../subdir':
  - app.sg
Tome name: subdir
Error loading '.../subdir/app.sg': [R0003] Runtime error: undefined variable: `greet`
```

**Control:** move that identical file to `subdir/greet.sg` (flat, same
directory) and use `invoke tome·greet·{greet};` — prints `from-subdir`. So the
module itself is fine; only its location is the problem.

Reproduced on `develop` @ `c6ef995`.

## Why it matters

Found while decomposing morgoth (`Daemoniorum-LLC/morgoth`) — a ~4,600-line
program now split into 18 sibling modules in one flat `src/`.

Five of those modules (`grid`, `vterm`, `terminal`, `layout`, `copy_mode`,
~1,600 lines) are a general-purpose terminal/TUI core with zero
application-specific coupling — the beginnings of a terminal counterpart to
Qliphoth. The natural next step is to group them under `src/tui/` so the seam
between the reusable layer and the application is visible in the tree, and so
the layer can later be lifted into its own tome.

That is not currently expressible. The alternatives are all worse: keep every
module flat and rely on naming conventions (`tui_grid.sg`, `tui_vterm.sg`), or
split into a separate tome before the API has settled.

## Notes

Whether the fix is nested tome paths (`tome·tui·grid`), flat resolution with
recursive discovery, or an explicit re-export module, some form of intra-tome
grouping seems needed before Sigil programs of this size are comfortable.

Filing as a gap rather than a defect — this may be an intentional design
constraint, in which case the useful outcome is documenting it, since the
failure mode (`undefined variable`) gives no hint that nesting is the problem.

---

paraphilic-ecchymosis:
Closes #109 (does not auto-close — this targets `develop`).

## Not an intentional constraint

The issue asks whether flat-only is by design. It is not: the **workspace
loader has always supported nesting**. `main.rs`, loading a crate's `src/`:

```rust
// e.g., "analyze.sigil" -> module name "analyze"
//       "router/mod.sigil" -> module name "router"
//       "router/types.sigil" -> module name "router·types"
```

`run` and `run-dir` simply never implemented it. `invoke tome·tui·grid·{Region}`
took the module name from the **second** segment alone:

```rust
let module_name = if prefix.len() >= 2 { prefix[1].clone() } else { simple_name.clone() };
let module_file_sg = format!("{}/{}.sg", source_dir, module_name);
```

so it looked for `tui.sg`, found nothing, and left the import unresolved —
surfacing as `undefined variable: Region`, which points nowhere near the
nesting.

## The rule

A path's segments do not say where the module stops and the imported item
begins — these are all the same shape:

```sigil
invoke tome·analyze;              // analyze.sg
invoke tome·output·Output;        // output.sg,   item Output
invoke tome·tui·grid·{Region};    // tui/grid.sg, item Region
invoke tome·tui·{layer_name};     // tui/mod.sg,  item layer_name
```

So `resolve_tome_module` drops the `tome` root, appends the final name, and
tries the **longest** module path first, giving up a segment at a time. At each
length it looks for `<path>.sigil`, `<path>.sg`, `<path>/mod.sigil`,
`<path>/mod.sg`, and the first that names a file wins. The module registers
under the `·`-joined form — `tui/grid.sg` is `tui·grid` — matching the workspace
loader.

`scroll foo;` was routed through the same resolver, so a module may be a file
**or** a directory carrying a `mod.sg`.

## Verified against the shape morgoth needs

```
nest2/
├── main.sg
├── app.sg
└── tui/
    ├── mod.sg
    └── grid.sg
```

```sigil
// tui/grid.sg
☉ Σ Region { ☉ w: i32, ☉ h: i32 }
⊢ Region { ☉ rite new(w: i32, h: i32) -> Self { Self { w, h } } }
☉ rite area(r: Region) -> i32 { r·w * r·h }

// app.sg
invoke tome·tui·grid·{Region, area};
☉ rite describe() -> i32 { area(Region·new(4, 5)) }

// main.sg
invoke tome·app·{describe};
invoke tome·tui·{layer_name};
☉ rite main() -> !i32 { println(layer_name()); println(describe()); ⤺ 0; }
```

On `develop` @ `c6ef995`: `undefined variable`. Here, under both `sigil run
./main.sg` and `sigil run-dir .`:

```
tui
20
```

Types and free functions both cross the directory boundary.

## Compatibility

Longest-first means `a/b.sg` beats `a.sg` for `tome·a·b·{…}` when both exist —
which is what the deeper spelling asks for. A flat `output.sg` is still found
for `tome·output·{…}`, the spelling every existing tome uses;
`a_flat_module_still_resolves_beside_a_directory_of_the_same_name` covers it,
and the 797-test Jormungandr suite (all flat tomes) is unchanged.

## What is still flat

`run-dir` lists and eagerly loads only the top level of the directory it is
given. Modules in subdirectories load on demand, when an `invoke` or `scroll`
reaches them — which is the Rust rule, and means an unreferenced file is not
executed. Making `run-dir`'s eager listing recurse touches the same function as
#119 (#107), so it belongs there rather than here.

## Tests

Rebased onto `develop` after #116 merged (5df25eb), so the baseline moved from
569 to 571 lib tests.

| | develop @ 5df25eb | this branch |
|---|---|---|
| `RUST_MIN_STACK=33554432 cargo test --release --no-fail-fast` (lib) | 571 passed / 2 failed | **575 passed / 2 failed** |
| bin | 41 passed / 0 failed | 41 passed / 0 failed |
| `jormungandr/tests/run_tests_rust.sh` | 797 / 4 failed / 2 skipped | 797 / 4 failed / 2 skipped |

Four tests added, three of them end-to-end against a tome written to a temp
directory (module resolution reads the filesystem, so a source string cannot
exercise it). The two nesting cases were confirmed failing on a `develop` build
before the change — with exactly the `undefined variable` the issue reports.

`docs/findings/tome-module-paths-and-subdirectories.md` records the resolution
order and the four spellings.

Rebased on top of #116 (#104), and independent of #119 (#107): #119 touches `run_directory` in
`main.rs`, this touches module resolution in `interpreter.rs`. #116 and this one
both touch `interpreter.rs` but in unrelated regions.

### DIFF
--- a/docs/findings/tome-module-paths-and-subdirectories.md
+++ b/docs/findings/tome-module-paths-and-subdirectories.md
@@ -1,0 +1,51 @@
+# How `invoke tome·…` finds a module file
+
+**Recorded:** 2026-09-10, closing #109.
+
+A tome used to have to be one flat directory. `invoke tome·tui·grid·{Region}`
+took the module name from the **second** segment alone, looked for `tui.sg`,
+and failed with `undefined variable: Region` — which points nowhere near the
+nesting.
+
+Subdirectories were never meant to be unreachable: the workspace loader
+(`main.rs`) has always read `router/types.sigil` as the module `router·types`
+and `router/mod.sigil` as `router`. `run` and `run-dir` simply did not.
+
+## The rule
+
+A path's segments do not say where the module stops and the imported item
+begins:
+
+```sigil
+invoke tome·analyze;              // analyze.sg
+invoke tome·output·Output;        // output.sg,      item Output
+invoke tome·tui·grid·{Region};    // tui/grid.sg,    item Region
+invoke tome·tui·{layer_name};     // tui/mod.sg,     item layer_name
+```
+
+So `resolve_tome_module` drops the `tome` root, appends the final name, and
+tries the **longest** module path first, giving up one segment at a time. For
+each candidate length it looks for, in order:
+
+```
+<path>.sigil    <path>.sg    <path>/mod.sigil    <path>/mod.sg
+```
+
+The first that names a file wins, and the module is registered under the
+`·`-joined form — `tui/grid.sg` is the module `tui·grid`, matching the
+workspace loader.
+
+Longest-first means `a/b.sg` beats `a.sg` for `tome·a·b·{…}` when both exist.
+A flat `output.sg` is still found for `tome·output·{…}`, which is the spelling
+every existing tome uses.
+
+`scroll foo;` goes through the same resolver, so a module may be a file or a
+directory with a `mod.sg` in it.
+
+## What is still flat
+
+`run-dir` lists and eagerly loads only the top level of the directory it is
+given. Modules in subdirectories load on demand, when an `invoke` or `scroll`
+reaches them — which is the Rust rule, and means an unreferenced file is not
+executed. Making `run-dir`'s eager listing recurse belongs with the load-order
+rework in #107.
--- a/parser/src/interpreter.rs
+++ b/parser/src/interpreter.rs
@@ -3830,17 +3830,18 @@ impl Interpreter {
                         }
                     }
                 } else {
-                    // External module: mod foo; - try to load foo.sigil or foo.sg from same directory
+                    // External module: `scroll foo;` - foo.sigil, foo.sg, or a
+                    // directory of its own carrying foo/mod.sg, the way the
+                    // workspace loader already reads a subdirectory.
                     if let Some(ref source_dir) = self.current_source_dir {
-                        let module_path_sigil =
-                            std::path::Path::new(source_dir).join(format!("{}.sigil", module_name));
-                        let module_path_sg =
-                            std::path::Path::new(source_dir).join(format!("{}.sg", module_name));
-                        let module_path = if module_path_sigil.exists() {
-                            module_path_sigil
-                        } else {
-                            module_path_sg
-                        };
+                        let module_path = Self::resolve_tome_module(
+                            source_dir,
+                            std::slice::from_ref(&module_name),
+                        )
+                        .map(|(_, file)| std::path::PathBuf::from(file))
+                        .unwrap_or_else(|| {
+                            std::path::Path::new(source_dir).join(format!("{}.sg", module_name))
+                        });
 
                         if module_path.exists() {
                             crate::sigil_debug!(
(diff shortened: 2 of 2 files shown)


paraphilic-ecchymosis:
Verified independently. Built the branch and ran the exact nesting shape morgoth needs:

```
src/
├── main.sg          invoke tome·tui·grid·{Region}; invoke tome·tui·{layer_name};
└── tui/
    ├── mod.sg       layer_name() -> 7
    └── grid.sg      Region::new(w,h)
```

| | result |
|---|---|
| `develop` | `[R0000] Runtime error: no fie…` — fails to resolve |
| this branch | **exit 19** — i.e. `3*4 + 7`, correct |

(Worth noting how I nearly misread that: I looked at stdout, saw nothing, and briefly thought it had failed. The value is returned through the **process exit code**, which is the CLI behaving exactly as designed. Same lesson as everything else today — read what the command actually said, not the channel you expected it to say it on.)

Lib suite: **575 passed, 2 failed** — the two being the pre-existing `llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}`. Baseline was 571/2, so +4 are yours.

**The strongest argument in this PR is that it is not inventing a rule.** The workspace loader has always supported nesting — `router/types.sigil` -> `router·types` is in its own comment. `run` and `run-dir` simply never implemented it, taking the module name from the second segment alone and looking for `tui.sg`. So this makes the two loaders agree rather than adding a third behaviour, and the module registers under the `·`-joined form the workspace loader already uses. That is a much safer change than it first appears.

The disambiguation is also the right shape for the actual ambiguity: a path's segments do not say where the module stops and the item begins, so trying the **longest** module path first and giving up a segment at a time is the only rule that handles `tome·analyze`, `tome·output·Output` and `tome·tui·grid·{Region}` uniformly. Guessing from segment count could not.

This unblocks something concrete: morgoth's reusable TUI layer was described as "inexpressible today" because a `tui/` namespace could not exist. It can now.

Approving and merging.

Issue #105, opened 2026-09-10
Rescued: two 2026-02 stashes, incl. raw-pointer type unification in typeck


paraphilic-ecchymosis:
> [!WARNING]
> **Corrected 2026-09-10 — do not apply the remaining hunks verbatim.**
>
> The `typeck.rs` hunk turned out to fix a live defect (#120) and has landed in
> **#129** — but **not as written**: its mutability clause is inverted for this
> codebase's call convention, and this issue originally claimed the opposite.
> See [Correction](#correction) below.
>
> The natural next move, now that one hunk has proved out, is to apply the rest
> of the stash verbatim. **Don't.** The other hunks
> (`sigil_runtime.c`, `llvm_codegen.rs`) were written by the same author in the
> same sitting under the same assumptions, and have had **no more testing than
> the one hunk that turned out to be backwards**. One hunk being correct in
> substance is not evidence for the others; it took a purpose-built test to
> catch the defect in this one, and reading it was not enough.

Two `git stash` entries from **2026-02-20** existed only in one workstation's local reflog. They have been pushed as branches so they exist somewhere durable. Nothing merged; this issue is the index.

A stash is the most losable thing in git — it lives in `.git/refs/stash` and its reflog, is invisible to `git log`, `git branch` and every remote, and a `git stash drop` or a reflog expiry takes it silently.

| branch | from | content |
|---|---|---|
| `rescue/stash-0-2026-02` | `stash@{0}` | the substantive one, below |
| `rescue/stash-1-2026-02` | `stash@{1}` | a one-line `parser/Cargo.lock` bump — noise, safe to drop |

Both are still in the local stash list; nothing was moved or applied.

## What is in stash-0

`WIP on fix/dogfooding: 7e07e3a fix(stdlib): fix stdin poll and PTY slave echo for interactive terminals`

```
parser/runtime/sigil_runtime.c  +32
parser/src/llvm_codegen.rs      +22
parser/src/typeck.rs            +10
```

**`typeck.rs` — raw-pointer type unification.** This is the interesting part:

```rust
// Raw pointers: *const T == *const T, *mut T == *mut T
// Also allow *mut T -> *const T (weakening mutability, safe coercion)
(Type::Ptr { mutable: ma, inner: a }, Type::Ptr { mutable: mb, inner: b }) => {
    let mut_ok = ma == mb || (!mb); // *mut -> *const is ok, not the reverse
    mut_ok && self.unify(a, b)
}
[...]

PR #129, opened 2026-09-10
fix(typeck): unify identical raw pointer types across a call (#120) — lands the rescued #105 stash

### LINKED ISSUE #120: Identical `*Δ c_void` types fail to unify across a call; diagnostic prints "expected *mut c_void, found *mut c_void"
Found while migrating **infernum-sigil** with a release compiler built from `develop` @ `c6ef995`.

A `*Δ c_void` argument passed from one function to another that declares the *same* type is rejected, and the diagnostic prints **the same text on both sides**.

## Reproduction

Whole file, two lines:

```sigil
rite cap(s: *Δ c_void) -> Result<B!> { B·mk(s) }
rite m(s: *Δ c_void) -> Result<B!> { cap(s) }
```

```
sigil check repro.sigil
[E0003] Error: type mismatch in argument 1: expected *mut c_void, found *mut c_void
```

Expected and found render identically, so the message cannot be acted on as written.

One function alone is clean — it needs the call from the second:

```sigil
rite cap(s: *Δ c_void) -> Result<B!> { B·mk(s) }   // ✓ no errors
```

`B` and `B·mk` are deliberately undefined and that is not the cause; the same
error appears in the original file where every type involved is defined. It also
reproduces through `⊢` blocks:

```sigil
⊢ A { ☉ rite cap(s: *Δ c_void) -> Result<B!> { B·mk(s) } }
⊢ P { ☉ rite m(s: *Δ c_void) -> Result<Self!> { ≔ e! = A·cap(s)?; Ok(Self { x: 1 }) } }
```

Generics and `where` clauses are not required — I reduced them away; they were in
the original site but make no difference.

## Provenance

Reduced from `src/native/cuda_graph.sigil:373` in infernum-sigil (632 lines →
2), where it reads:

```sigil
GraphExec·from_graph(graph)
```

## Possibly related

#105 records a rescued 2026-02 stash described as "raw-pointer type unification
in typeck", on `rescue/stash-0-2026-02`. I have not applied or tested that
branch — flagging it in case it is the same defect, already half-fixed and
unmerged.

## Two defects or one

Worth separating when this is picked up:

1. The unification failure itself — two identical `*Δ c_void` types not unifying
   across a call boundary.
2. The diagnostic rendering both sides identically, which would be worth fixing
   regardless, since it makes any raw-pointer mismatch unreadable.

Filed from infernum-sigil; I do not own this repo, so filing rather than fixing.

paraphilic-ecchymosis:
Fixes #120. **The fix is the rescued 2026-02-20 stash from #105, not a reimplementation.**

## The defect

`unify` has **no `(Ptr, Ptr)` arm at all** on `develop`. Two identical
`*Δ c_void` types therefore fall through to the catch-all and fail, and the
diagnostic renders both sides from the same type:

```sigil
rite cap(s: *Δ c_void) -> Result<B!> { B·mk(s) }
rite m(s: *Δ c_void) -> Result<B!> { cap(s) }
```
```
[E0003] type mismatch in argument 1: expected *mut c_void, found *mut c_void
```

Found while migrating infernum-sigil; reduced from 632 lines to those two.

## Provenance — this is #105's stash landing

The typeck.rs hunk is taken from `rescue/stash-0-2026-02`, the stash indexed in
#105 that existed only in one workstation's local reflog until it was pushed.
Applied rather than rewritten, so the rescue is what lands and the seven-month-old
work gets the credit.

The stash's other hunks — `stdlib.rs`, `llvm_codegen.rs`, `sigil_runtime.c`,
`Cargo.lock` — are unrelated stdin/PTY work and are **not** included. This PR is
the pointer-unification hunk only.

## One correction to the stash

Its mutability-weakening clause is inverted for this codebase's call convention.
Arguments are checked as `self.unify(param, arg)`, so the first binding is the
declared parameter and the second the supplied argument. The stash tested `!mb`
— the *argument* — which permits a `*const` argument to silently gain
mutability, and rejects the safe case the comment describes.

```rust
let mut_ok = ma == mb || (!mb);   // stash: allows *const → *mut
let mut_ok = ma == mb || !ma;     // here:  allows *mut → *const
```

After the correction:

| argument | parameter | |
|---|---|---|
| `*mut u8` | `*const u8` | accepted — weakening, safe |
| `*const u8` | `*mut u8` | rejected — would gain mutability |

I did not spot this by reading; the spec test added here caught it, which is why
both directions are in the test.

## Verification

- The #120 reproduction checks clean here and fails on `develop`.
- `src/native/cuda_graph.sigil` in infernum-sigil — where this surfaced — no
  longer reports a pointer error, and no `c_void` mismatch remains anywhere in
  that 187-file tree.
- `sigil check` across all **724** files in `jormungandr/tests/spec`: **zero
  regressions**, both before and after. The only newly-passing file is the test
  added here.

## Note for whoever reviews

#120 also flagged the diagnostic itself — rendering expected and found
identically makes any raw-pointer mismatch unreadable. This PR removes the
*cause* in the identical-type case but does not change the formatter, so a
genuine mismatch between two distinct pointer types will still print two
identical-looking strings if their `Display` collides. Worth a separate look.

### DIFF
--- a/jormungandr/tests/spec/08_ffi/P1_008_raw_pointer_unification.expected
+++ b/jormungandr/tests/spec/08_ffi/P1_008_raw_pointer_unification.expected
@@ -1,0 +1,1 @@
+raw pointer unification ok
--- a/jormungandr/tests/spec/08_ffi/P1_008_raw_pointer_unification.sg
+++ b/jormungandr/tests/spec/08_ffi/P1_008_raw_pointer_unification.sg
@@ -1,0 +1,21 @@
+// Test: identical raw pointer types unify across a call boundary
+// Spec: 08-FFI.md § 2.1 Pointer Types
+// Priority: P1 - Standard Feature
+// Regression: #120 -- `*Δ c_void` passed to a parameter of the same declared
+// type was rejected with "expected *mut c_void, found *mut c_void".
+
+rite takes_mut(p: *Δ u8) -> i32 { 0 }
+rite takes_const(p: *const u8) -> i32 { 0 }
+
+rite forward_mut(p: *Δ u8) -> i32 {
+    takes_mut(p)
+}
+
+// *mut → *const weakens mutability and is allowed.
+rite mut_to_const(p: *Δ u8) -> i32 {
+    takes_const(p)
+}
+
+rite main() {
+    println("raw pointer unification ok");
+}
--- a/parser/src/typeck.rs
+++ b/parser/src/typeck.rs
@@ -4238,6 +4238,22 @@ impl TypeChecker {
                 a.iter().zip(b.iter()).all(|(x, y)| self.unify(x, y))
             }
 
+            // Raw pointers: *const T == *const T, *mut T == *mut T
+            //
+            // Also allow *mut T → *const T (weakening mutability, safe coercion).
+            // `unify` is called as `unify(param, arg)`, so `ma` is the declared
+            // parameter and `mb` the supplied argument: the safe direction is a
+            // `*mut` argument reaching a `*const` parameter, which is `!ma`.
+            // Testing `!mb` instead permits the opposite -- a `*const` argument
+            // silently gaining mutability -- and rejects the safe case.
+            (Type::Ptr { mutable: ma, inner: a }, Type::Ptr { mutable: mb, inner: b }) => {
+                let mut_ok = ma == mb || !ma;
+                mut_ok && self.unify(a, b)
+            }
+            // Allow *const T / *mut T to coerce with &T (auto-deref context)
+            (Type::Ptr { inner: a, .. }, Type::Ref { inner: b, .. }) => self.unify(a, b),
+            (Type::Ref { inner: a, .. }, Type::Ptr { inner: b, .. }) => self.unify(a, b),
+
             // References
             (Type::Ref { mutable: ma, inner: a, .. }, Type::Ref { mutable: mb, inner: b, .. }) => {
                 // Allow &[T; N] to coerce to &[T] (array to slice)


paraphilic-ecchymosis:
Verified independently, and I specifically checked the direction that matters for safety rather than only the one the fix enables.

| case | develop | this branch |
|---|---|---|
| `*mut → *const` (weakening, safe) | ✗ rejected | ✓ accepted |
| `*mut → *mut` (identical) | ✗ rejected | ✓ accepted |
| **`*const → *mut` (gaining mutability, unsafe)** | — | **✓ correctly REFUSED**, exit 1, `expected *mut U8, found *const U8` |

That third row is the one I wanted to see fail, and it does. Lib suite 571 passed / 2 failed — the two pre-existing `test_generic_struct_*`.

**This is the rescued stash landing, and it is better than the stash was.**

That code existed only in a local `git stash` reflog until this morning — invisible to `git log`, `git branch` and every remote, one `stash drop` from gone. It is now `rescue/stash-0-2026-02` and indexed in #105.

**And you caught that the stash was wrong in the direction that matters.** I want that on the record because I got it wrong first: when I indexed the stash I wrote that it "handled the mutability direction correctly — `*mut → *const`, not the reverse". I was reading its **comment**, not its semantics under this codebase's call convention. Arguments are checked as `unify(param, arg)`, so the stash's `!mb` tests the *argument* — permitting a `*const` argument to silently gain mutability, and rejecting the safe case its own comment described. Applying it verbatim would have shipped a pointer-safety hole with a comment claiming the opposite.

The code comment you wrote in its place is the right artefact: it states the call convention, why `!ma` is correct, and what `!mb` would wrongly permit. Someone will otherwise "simplify" that back.

Your note that the stash's **remaining** hunks were written under the same assumptions and have had no more testing than this one is the most valuable line in #105, because the natural move now that one hunk proved out is to apply the rest verbatim.

Approving and merging.

PR #130, opened 2026-09-10
fix(cli): gate wasm and rust codegen on the typecheck, like llvm and compile

### LINKED ISSUE #114: `sigil wasm` skips the pre-codegen typecheck #96 added — emits a valid module for a program `check` rejects
#96 made `sigil llvm` and `sigil compile` type-check before codegen, so they stop instead of emitting broken IR. **`sigil wasm` was not included in that gate**, and still emits a module for a program the checker rejects.

Found while re-measuring the Lares capability probe against `develop` @ `c6ef995`, compiler built `--no-default-features --features minimal,protocols,react-migrate,wasm`.

## Reproduction

`src/lib.sigil`:

```sigil
☉ rite bad(x: f64) -> i64! {
    ⎇ x ≥ 70 { 1 } ⎉ { 0 }
}
rite main() { println("ok"); }
```

```
$ sigil check src/lib.sigil            ; echo $?
[E0003] Error: comparison operands must have same type: left=Float(F64), right=Int(I64)
1

$ sigil compile src/lib.sigil -o native ; echo $?      # gated by #96
1

$ sigil wasm src/lib.sigil -o one.wasm  ; echo $?      # NOT gated
0
```

Exit 0, and a 4,472-byte module lands. Same for the project form, `sigil wasm <dir>/ -o out.wasm`. Deterministic, 3/3 runs each.

The module is not merely emitted, it is **structurally valid** — it instantiates:

```
$ node -e 'new WebAssembly.Module(require("fs").readFileSync("one.wasm")); …'
WebAssembly.Module: VALID, 4472 bytes
```

So the failure mode is not a crash at instantiate. It is a program that loads and runs, with the backend having silently picked some meaning for a comparison the type checker says is meaningless.

## Why this one is worth prioritising

WASM is the only backend the Qliphoth/Lares port has ever run on. The whole validation chain for that work is:

1. `sigil check` each file
2. `sigil wasm project/ -o project.wasm`  ← claimed to be the step that catches what (1) cannot
3. render and survey against the built module

Step 2 is documented in the Lares handoff as the guard against a link failure that per-file checking misses. It cannot serve that role while it is the one command that does not type-check, and the survey in step 3 reads the module step 2 produced — so a type error propagates through the entire chain as a pass.

Concretely, in the current Lares client: `project/src/score_bar.sigil` fails `sigil check` with this exact E0003 (sigil-lang#112), and `sigil wasm project/` still returns **exit 0** and produces a 362KB module containing it.

## Suggested fix

Whatever #96 added ahead of the LLVM and native codegen entry points, add ahead of the WASM one — including the directory/project path, which reaches codegen separately from the single-file path (both are shown failing to gate above).

Worth checking `sigil rust` and `sigil jit` in the same pass; I did not test those.

paraphilic-ecchymosis:
Closes #114. (`Closes` does not auto-close here — I will close it by hand once this merges.)

#96 made `sigil llvm` and `sigil compile` type-check before codegen. `wasm` and `rust` were left out:

| | `check` | `run` | `jit` | `llvm` | `compile` | `wasm` | `rust` |
|---|---|---|---|---|---|---|---|
| before | 1 | 1 | 1 | 1 | 1 | **0** | **0** |
| after | 1 | 1 | 1 | 1 | 1 | **1** | **1** |

`sigil wasm` did not merely emit *something* for a program `sigil check` rejects — it emitted a **structurally valid module that instantiates**. Nothing crashed, nothing warned, and the backend had silently picked a meaning for a comparison the type checker calls meaningless.

That mattered beyond the two commands. The Lares/Qliphoth probe used `sigil wasm project/` as the step that catches what per-file `sigil check` cannot see, and then surveyed the module that step produced — so a type error travelled the whole validation chain as a pass, and the chain's own design document named that step as the guard.

## What changed

`typecheck_before_codegen` was `#[cfg(feature = "llvm")]`, so it did not exist in a wasm-only build. It is now unconditional and reached from three more places: the single-file wasm path, the wasm project path, and `rust_compile_file`.

## The project path, and the version of it I threw away

Worth recording, because it looks equivalent and is not.

My first attempt walked `<project>/src` and checked everything under it. But a project build enters through `lib.sigil` and follows module resolution, so `src/` is a **superset** of what it compiles. Measured against five bundled projects, that version rejected four of them — over files that **do not even parse**, because they are unported Rust sitting beside the real sources and the build had never looked at them.

A gate that fails builds over files codegen ignores is not a stricter version of this check. It is a different bug, and it would have been a worse one than the thing I was fixing.

So the gate asks the compiler which modules it actually loaded (`WasmCompiler::loaded_modules`, added here) and checks exactly those. Resolution has to run to answer that, so for the project path the check sits after compilation and **before `fs::write`** — no artifact reaches disk for a program the checker rejects, which is the property the issue was actually about.

Both new gates report every failing module rather than stopping at the first, and sort the file set. A `HashSet` has no order, and diagnostics that reshuffle between runs on identical input are diagnostics nobody can trust.

## Validation

| | |
|---|---|
| exit-code matrix | all seven commands return 1; `sigil wasm <dir>` too |
| no stale artifact | the output file is **absent** after a failed build, where before a module was written |
| blast radius | five bundled projects: `sigil-web` and `codec` still build and still write modules. `qliphoth-sys`, `qliphoth-router` and `website-qliphoth` fail — and **failed before this change**, at parse time in the pre-existing `Err` arm this patch does not touch |
| lares-client | `sigil wasm project/` now exits 1 naming exactly one file, `score_bar.sigil`, which fails `sigil check` via #112 |
| `cargo test --lib` | **757 passed, 0 failed** |
| Sigil suite | **798 passed, 4 failed** — `P0_005_kafka_producer`, `P0_007_amqp`, `P1_006_kafka_consumer`, `P1_008_amqp_consumer`, all `Runtime error` from absent brokers |

The lares-client result is the gate working, not a regression: the module it used to emit was for a program the checker rejects. Anyone re-measuring that repo should expect the link step to go from exit 0 to exit 1 until #112 lands.

Counts are under `--features minimal,protocols,wasm`, since default features need LLVM 18. Worth a CI confirmation under the full set rather than taking these as covering it — in particular I could not exercise the `llvm`/`compile` paths locally, and this patch widens a `cfg` they depend on.

## A gap this leaves open

The negative case pins the **type checker's verdict** on the reproduction. It does not pin the **exit codes**, because the suite runs programs through the interpreter and has no way to ask what `sigil wasm` returned.

That absence is why #114 survived #96 in the first place: the same fix was made for two commands, and nothing in the suite could ask the other two. A CLI-level harness that asserts every backend agrees on a rejected program would have caught it the day #96 landed, and would catch the next backend added. I have not built one here — it is a larger change than this fix and wants its own discussion — but it is the thing that actually prevents a recurrence.

## Do not review this independently of #137

**This PR and [#137](https://github.com/Daemoniorum-LLC/sigil-lang/pull/137) are coupled, and the coupling is invisible from here.**

The negative case this PR adds originally used the f64/int comparison from #114's reproduction. #137 makes that comparison legal — the checker already coerced int→float in argument and arithmetic position, and comparison refusing it was the real defect behind #112.

So had both merged as written, this test would have kept passing **while testing nothing**. Not failing loudly: passing, silently, forever.

Already handled — the case now uses bool-against-int, which has no coercion path in any position, and I verified it exits 1 under builds of *both* branches. No action is needed. It is stated here because a reviewer who opens this PR first has no way to know #137 exists, and the failure mode is a green light rather than a red one.

That is sigil-lang#131 occurring inside my own work: a check that quietly changed the question it was answering.

## Coordination

Found by the Lares capability probe (gap **S128**). Claimed `parser/src/main.rs` on the Lares coord-log before starting, and flagged there that the shared `sigil-lang` checkout has uncommitted edits to the same file in `llvm_file`/`compile_file` — different functions from the ones here, so this is a merge-conflict risk rather than a clobber risk. Work done in a separate worktree off `origin/develop`; no other checkout was touched.


paraphilic-ecchymosis:
Verified, after first running a check that answered the wrong question — which is worth recording since it is this repo's house pattern.

My first attempt was `sigil wasm` on ill-typed input. Both builds exited 1, so it looked like no change. They were both refusing because the binary lacks `--features wasm` — the check answered "is wasm compiled in?" while I read it as "does the gate work?". Exactly #131, committed by the person reviewing the fix for it.

The `rust` path needs no feature flag, so:

```sigil
rite add(a: i64, b: i64) -> i64 { a + b }
rite main() { ⤺ add(1); }
```

| | result |
|---|---|
| `develop` | **exit 0** — transpiles ill-typed code silently |
| this branch | **exit 1** — `Type error in 'bad.sigil': expected 2 arguments, found 1` |

Lib suite 571 passed / 2 failed, the two pre-existing `test_generic_struct_*`.

This closes the last hole #96 left. `sigil llvm` and `sigil compile` got the pre-codegen typecheck then; `wasm` and `rust` did not, so two of four codegen paths still turned ill-typed input into output rather than a diagnostic. Emitting a *valid-looking* Rust file from code the checker rejects is the worse half of that, because the failure surfaces later in someone else's toolchain with no trace back to here.

Merged.

Issue #92, opened 2026-09-10
migrate rewrites source files in place — it needs an output destination


paraphilic-ecchymosis:
`sigil migrate <file|dir>` converts Rust syntax to Sigil by **overwriting the input**. There is no output flag, no dry-run, and no backup.

```
$ cp somecrate/src/lib.rs /tmp/t.rs
$ sigil migrate /tmp/t.rs
$ cmp somecrate/src/lib.rs /tmp/t.rs
... differ: byte 196, line 3
```

That is destructive by default, and the blast radius is as large as whatever path you hand it. Pointing it at a checkout rewrites the checkout; pointing it at `~/.cargo/registry` corrupts the local crate cache for every project on the machine. Nothing warns you first.

It is also bad UX for the tool's actual job. Migration is something you want to *inspect* — run it, read the diff, decide. Right now the only way to see what it would do is to copy the tree first and diff afterwards, which every caller has to reinvent. Two sessions have already had to build that scaffolding by hand just to check the reserved-name diagnostic for regressions (see #90).

## What it should do

- Write to a destination: `sigil migrate <src> -o <dest>`, mirroring the input tree.
- Default to **not** modifying the input. In-place should be opt-in and explicit — `--in-place` — not the only mode.
- A `--dry-run` / `--diff` that reports what would change without writing anything. This is also what makes #90's sweep harness cheap to build.

## Why now

The reserved-name diagnostic (#68) prints its warnings as a side effect of the same command that rewrites the file. So the *only* way to run the diagnostic is to destroy the input. A read-only analysis path is worth having on its own, independent of the migration.

Reported by Daemon: "The migrate absolutely should have an output directly, not write in place. That's destructive and a poor UX."

Issue #91, opened 2026-09-10
migrate: `|mut x|` closure parameters are still unjudged


paraphilic-ecchymosis:
Follow-up to #89, which fixed closure-parameter classification generally but leaves one shape uncovered.

A closure parameter is found either by the colon in `|x: T|`, or — for the untyped form — by sitting between the opening bar (or a comma) and the next comma or closing bar. `|mut tome|` satisfies neither: the byte before `tome` is the `t` of `mut`, so the untyped test fails, and there is no colon.

```rust
let f = |mut tome| tome;   // not judged
let f = |mut tome: u32| tome;   // judged — the colon finds it
```

Quiet false negative, in the same direction as #84: a name that is fine as a field and refused as a binding goes unwarned.

## Size

Small. The untyped test in `collect_in_code` is:

```rust
let untyped_closure_param = in_closure_params
    && matches!(before, b'|' | b',')
    && matches!(rest.as_bytes().first(), Some(b',') | Some(b'|'));
```

The first clause needs to also accept a preceding `mut` whose own predecessor is the bar or a comma — `prev_words` already carries what is needed.

## Do not fix it without re-checking the sweep

The reason #84 sat open was a judgement that an untested heuristic in this scanner is a bad trade, and that judgement was sound: the bracket tracker has been the largest source of wrong answers here. Whatever fixes this must re-run a false-positive sweep. See #90 — there is no committed harness to do that with, which is the actual blocker.

Deliberately filed rather than left as a comment in the PR body: a live gap does not belong only in a merged diff.

paraphilic-ecchymosis:
Fixed by #135, merged into `develop`. Verified: bin suite 46 passed / 0 failed, and the four corpus tests from #133 still green.

This issue was **correctly blocked** and the block was worth respecting. #91 said `|mut tome|` should not land without a false-positive sweep, and that the sweep did not exist — which was #90. #133 built the harness, this landed against it, and the harness confirmed no regression on the four shapes the scanner has historically got wrong.

Worth noting because "blocked on another issue" is usually a way of not doing something: here it was load-bearing. Landing #91 first would have meant changing a scanner whose headline property — 0 false positives — nobody could re-check.

(Closing by hand — `Closes #91` only fires on merge into the default branch.)

PR #133, opened 2026-09-10
test(migrate): a harness for the reserved-name corpus claim

### LINKED ISSUE #90: migrate: no reusable harness backs the corpus claim behind the reserved-name diagnostic
The reserved-name diagnostic (#68) rests on a headline property: **zero false positives across a 6,091-file validation corpus**. That number is not reproducible from anything on the branch.

There is no harness in `tools/`, none in #68's commits, and none in its worktree. Re-checking the property today means reconstructing a sweep from scratch — which two separate sessions have now each done differently (21,897 registry files one way, a 2,500-file per-file sample the other).

That is backwards for the one claim the feature is sold on. Every change to the scanner — #84/#89 was the second — has to re-establish the property by hand, and two hand-built sweeps are not comparable to each other or to the original 6,091.

## What this wants

- A committed harness that walks a defined corpus and reports findings in a stable, diffable form.
- The corpus **defined** rather than implied: "the crate registry" is not reproducible across machines or months. Pin it, or make the harness report what it actually scanned.
- Cheap enough to run per-PR, so "0 false positives" is a check rather than a memory.

## Trap to encode in it

`sigil migrate` **rewrites files in place**. Any sweep must copy first, or work through `collect_keyword_collisions` directly. Pointing it at `~/.cargo/registry` corrupts the cargo cache.

## Related

- #68 — the diagnostic
- #84 / #89 — the closure-parameter fix, where develop turned out to carry a live false positive (`this` as a closure parameter in `cranelift-isle` and `rustc-demangle`) that the original corpus run either never covered or never surfaced

paraphilic-ecchymosis:
Closes #90 (does not auto-close — this targets `develop`).

Two things, answering different questions. The issue asks for a committed
harness, a *defined* corpus, and something cheap enough to run per PR — those
are not the same artefact, so this is two.

## The per-PR check — `tools/collision-sweep/corpus/`

Four files, one per shape the scanner has been wrong about, with their findings
pinned **exactly** in `collision_corpus_tests` (`parser/src/main.rs`).
Milliseconds, inside `cargo test`.

| fixture | shape | why it is there |
|---|---|---|
| `positions.rs` | one name in two positions | `tome` is a valid field and an invalid binding; `this` is the reverse — a scanner with one verdict per word fails this whichever way it decides |
| `closure_params.rs` | closure parameters | #84/#89 — develop carried `\|this\|` as a **live** false positive |
| `prose.rs` | colons in comments | "Implementation based on:" is not a field |
| `bounds.rs` | `where Self: Sized` | reads exactly like a field declaration |

Expectations are the literal finding lists, e.g.

```rust
expect("positions.rs", &[("this", 1), ("tome", 1)]);
expect("prose.rs", &[]);
```

A scanner change that alters one has to change it in the same commit, where a
reviewer sees it. That is what turns the property from a memory into a check.

## The wide measurement — `sweep.sh` / `compare.sh`

```console
$ ./sweep.sh ~/.cargo/registry/src -o /tmp/sweep-before
sweep: 31464 files under /home/…/.cargo/registry/src
…
files with findings: 76
findings (rows):     80
distinct names:      12
```

Output is `manifest.txt` (root, file count, **digest of the sorted relative
path list**, binary), `findings.tsv` (`path` ⇥ `name` ⇥ `uses`, sorted),
`summary.txt`, and the `raw.err` it was parsed from.

The corpus is **reported, not assumed** — the issue's point that "the crate
registry" is not reproducible across machines or months is exactly right, so
the harness records what it actually scanned. `compare.sh` warns loudly when
two sweeps' digests differ, and otherwise diffs findings, exiting 1 on any new
row:

```console
$ ./compare.sh /tmp/sweep-before /tmp/sweep-after
before: 4 findings
after:  5 findings
added:   1
removed: 0

ADDED (each of these needs an argument):
  + closure_params.rs	tome	2
```

Every added row is either the true positive a change set out to find or the
regression it did not. Rows that disappear are fixed false positives.

## What it measures today

develop @ `5df25eb`, 31,464 files under `~/.cargo/registry/src`: **76 files
with findings, 80 rows, 12 distinct names** (`aspect` 29, `simd` 20, `this` 12,
`ref` 6, `of` 4, then a tail).

I spot-checked all 12 `this` findings, since `this` is the name #84/#89 was
about. Every one is a **struct field**:

```rust
this: &'a mut ThinArc<H, T>,   // triomphe-0.1.15/src/thin_arc.rs
this: V2,                      // pulp-0.22.2/src/x86/v2.rs
this: &'a TokenStream,         // logos-codegen-0.14.4/src/generator/mod.rs
```

`this` is refused as a field and accepted as a binding, so these are true
positives — the closure-parameter false positive #89 fixed is gone, and what
remains under that name is correct. That is the kind of claim that was not
checkable before.

## The trap, encoded

`sweep.sh` only ever runs `--dry-run`, which since #92/#94 writes nothing
anywhere — that is what makes it safe to point at `~/.cargo/registry`. The
README says not to add a writing mode, and the script has no flag that could
select one.

## Tests

Baseline measured on `develop` (5df25eb) first, same machine, same flags:

| | develop | this branch |
|---|---|---|
| `RUST_MIN_STACK=33554432 cargo test --release --no-fail-fast` (lib) | 571 passed / 2 failed | 571 passed / 2 failed |
| bin | 41 passed / 0 failed | **45 passed / 0 failed** |

The 2 failures are the pre-existing

This is the blocker #91 names. #91 follows on top of this branch, with a
before/after sweep as its evidence.

### DIFF
--- a/parser/src/main.rs
+++ b/parser/src/main.rs
@@ -8483,3 +8483,71 @@ mod migrate_destination_tests {
         assert_eq!(fs::read_to_string(&src).unwrap(), RUST_SOURCE);
     }
 }
+
+/// The committed corpus behind the reserved-name diagnostic.
+///
+/// #68 is sold on a corpus property -- "0 false positives across N files" --
+/// that nothing on the branch reproduced, so every change to the scanner
+/// re-established it by hand and no two hand-built sweeps were comparable
+/// (#90). `tools/collision-sweep/sweep.sh` is the wide measurement; this is
+/// the per-PR check: four files, one per shape the scanner has been wrong
+/// about, with their findings pinned exactly.
+///
+/// A scanner change that alters any expectation below has to change it here,
+/// in the same commit, where a reviewer sees it.
+#[cfg(test)]
+mod collision_corpus_tests {
+    use super::collect_keyword_collisions;
+
+    fn findings(fixture: &str) -> Vec<(String, usize)> {
+        let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("../tools/collision-sweep/corpus")
+            .join(fixture);
+        let source = std::fs::read_to_string(&path)
+            .unwrap_or_else(|e| panic!("read {}: {e}", path.display()));
+        collect_keyword_collisions(&source)
+    }
+
+    fn expect(fixture: &str, want: &[(&str, usize)]) {
+        let got = findings(fixture);
+        let got: Vec<(&str, usize)> = got.iter().map(|(n, c)| (n.as_str(), *c)).collect();
+        assert_eq!(got, want, "findings for {fixture}");
+    }
+
+    #[test]
+    fn a_name_is_judged_by_its_position_not_by_itself() {
+        // `tome` is a valid Sigil field and an invalid binding; `this` is the
+        // other way round. Both appear in both positions in the fixture, so a
+        // scanner with one verdict per word fails this whichever way it
+        // decides: only the parameter `tome` and the field `this` are refused.
+        expect("positions.rs", &[("this", 1), ("tome", 1)]);
+    }
+
+    #[test]
+    fn a_closure_parameter_is_a_binding() {
+        // #84/#89. `this` binds cleanly, so no closure parameter spelled
+        // `this` may be reported -- develop carried exactly that as a live
+        // false positive, in cranelift-isle and rustc-demangle. `of` is
+        // refused in every position, so it is a true positive and stays.
+        //
+        // `tome` is counted once: the typed `|mut tome: u32|` is found by its
+        // colon. The untyped `|mut tome|` beside it is #91 -- a quiet false
+        // negative, in the same direction as #84.
+        expect("closure_params.rs", &[("of", 1), ("tome", 1)]);
+    }
+
+    #[test]
+    fn prose_is_not_code() {
+        // Prose ends a sentence with a colon too. Reading comments reported
+        // every colon-terminated word in a doc comment as a field.
+        expect("prose.rs", &[]);
+    }
+
+    #[test]
+    fn a_where_bound_is_not_a_field() {
+        // `where Self: Sized` inside a trait body reads exactly like a field
+        // declaration.
+        expect("bounds.rs", &[]);
+    }
+}
+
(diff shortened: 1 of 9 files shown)


paraphilic-ecchymosis:
Pushed one refinement: the manifest now records the binary by **sha256** rather than by path and timestamp.

`sigil` has no `--version`, so the field read `unknown` — and a path plus a timestamp does not say which code ran, which is the whole question a before/after comparison is asking. `compare.sh` now also notes when two sweeps used the same binary, since comparing a build against itself is not what anyone runs it for.

```
root:         …/corpus
files:        4
path-digest:  5e3c27d4120cba2b8d86488c1dd6ffc0ac26983128e696bd998ba36f8d9ec4f5
binary:       …/parser/target/release/sigil
binary-sha256: ac229db6057c083e2611f1d984ec06a6bf42fbff23f24ac112af59c5ffb2bfc3
swept:        2026-09-10T11:37:39Z
```

Force-pushed as `f70fdcd`; #135 rebased onto it.

paraphilic-ecchymosis:
Verified: the four corpus tests run and pass in milliseconds inside `cargo test` (exit 0), which is the property #90 actually needed — a check cheap enough that nobody skips it.

```
collision_corpus_tests::a_name_is_judged_by_its_position_not_by_itself ... ok
collision_corpus_tests::a_closure_parameter_is_a_binding ... ok
collision_corpus_tests::prose_is_not_code ... ok
collision_corpus_tests::a_where_bound_is_not_a_field ... ok
```

**Splitting this into two artefacts was the right read of the issue, and better than what I asked for.** #90 conflated "a committed harness", "a defined corpus" and "cheap enough per PR" as one thing; they answer different questions and one artefact would have served none of them well.

The fixture selection is the part I would highlight to anyone reviewing this later: **one file per shape the scanner has actually been wrong about**, not a sample of arbitrary code. `positions.rs` in particular encodes the thing that makes this scanner hard — `tome` is a valid field and an invalid binding, `this` is the reverse, so a scanner with one verdict per word fails whichever way it decides. That is the defect class from #84 pinned as an executable statement rather than prose.

`closure_params.rs` is worth its place too: develop carried `|this|` as a **live false positive**, and this repo would not have known. Now it would.

Context for the record, since #90 was mine and was written from a bad premise: I filed it saying no harness backed the "0 false positives across 6,091 files" claim, which was true. What I did not say, because I had not realised it, was that the claim itself had never been re-checkable — so every change to the scanner since had been validated by whoever happened to reconstruct a sweep that day, differently each time. Two sessions did exactly that within hours of each other. This ends that.

Approving and merging.

PR #135, opened 2026-09-10
fix(migrate): judge a closure parameter written `|mut x|`

### LINKED ISSUE #91: migrate: `|mut x|` closure parameters are still unjudged

paraphilic-ecchymosis:
Closes #91.

#133 (#90) has landed, and this is retargeted to `develop` — it was stacked on
that branch because #91's own text says the blocker is that there is no harness
to re-check the sweep with. The harness commit is now in `develop`, so this is
one commit.

## The gap

A closure parameter is found either by the colon in `|x: T|` or, untyped, by
*starting* and *ending* one — the byte before it is the bar or a comma, and a
comma or the closing bar follows:

```rust
let untyped_closure_param = in_closure_params
    && matches!(before, b'|' | b',')
    && matches!(rest.as_bytes().first(), Some(b',') | Some(b'|'));
```

`|mut tome|` satisfies neither: the byte before `tome` is the `t` of `mut`, and
there is no colon. `|mut tome: u32|` was always caught — by its colon — which
is what made the untyped form a *quiet* false negative rather than a visible
one.

## The fix

`mut` reaches back exactly one word, so the test looks past it to the bar or
comma before *that*:

```rust
let starts_a_param = matches!(before, b'|' | b',')
    || (prev_words[1] == "mut" && matches!(*prev_word_before, b'|' | b','));
```

`prev_words` already carried the word; `prev_word_before` on `CollisionScan` is
the one byte it did not. The clause stays gated by `in_closure_params`, so
nothing outside a parameter list moves — a local is still judged by `let mut`,
not by this, and a test is included for that.

## The sweep #91 asked for

> Do not fix it without re-checking the sweep.

Both halves of the #133 harness, run either side of this commit.

**Committed corpus** — exactly the intended change and nothing else:

```console
$ ./compare.sh /tmp/corpus-before /tmp/corpus-after
removed (no longer reported):
  - closure_params.rs	tome	1
ADDED (each of these needs an argument):
  + closure_params.rs	tome	2
```

`tome` goes from 1 use to 2: the untyped `|mut tome|` beside the typed one is
now judged. The expectation in `collision_corpus_tests` changed with it, in
this commit.

**Wide sweep** — `~/.cargo/registry/src`, 31,464 files, identical path digest
either side, so directly comparable:

```console
$ ./compare.sh /tmp/sweep-before /tmp/sweep-final
before: 80 findings
after:  80 findings
added:   0
removed: 0

no new findings
```

**0 added, 0 removed.** No new false positives — and no new true positives in
real crates either, which is consistent with the note already on
`a_closure_parameter_is_a_binding_not_a_field` that a 21,897-file re-sweep
found no reserved name used as a closure parameter in this direction. That half
is now *measured and absent* rather than unmeasured, which is the difference
#90 was for.

## Tests

Baseline measured on `develop` (5df25eb) first, same machine, same flags:

| | develop | #133 | this branch |
|---|---|---|---|
| `cargo test --release` (lib) | 571 / 2 failed | 571 / 2 failed | 571 / 2 failed |
| bin | 41 / 0 | 45 / 0 | **46 / 0** |

The 2 failures are the pre-existing

`a_mut_closure_parameter_is_still_a_parameter` covers the untyped and typed
forms, the second-parameter position, `move |mut …|`, and the two directions it
must *not* move: `|x: mut u32|` (the word before `u32` is `mut`, but a colon
precedes that `mut`) and `|mut this|` (a binding the parser accepts — the #84/#89
direction).

### DIFF
--- a/parser/src/main.rs
+++ b/parser/src/main.rs
@@ -6000,6 +6000,10 @@ struct CollisionScan {
     /// The last non-whitespace byte, for `!` before a bracket and `$` before a
     /// word.
     prev_byte: u8,
+    /// The byte before the *previous* word, so a name behind a `mut` can still
+    /// be seen to start a closure parameter: in `|mut tome|` the byte before
+    /// `tome` is the `t` of `mut`, and the bar is one word further back.
+    prev_word_before: u8,
 }
 
 /// Tally names in one span of code that the parser refuses where they stand.
@@ -6014,6 +6018,7 @@ fn collect_in_code(
         prev_words,
         in_where,
         prev_byte,
+        prev_word_before,
     } = state;
     // Everything above carries over from the previous span deliberately: a
     // comment is not a syntactic break, so `let /* why */ x: T` is still a
@@ -6115,8 +6120,16 @@ fn collect_in_code(
             // parameter named after its own type; the byte before it is the
             // `:`, so it is not. `|x: &'a Tome|` fails the same test on the
             // `a` of the lifetime.
+            // `|mut tome|` starts a parameter one word further back: the byte
+            // before the name is the `t` of `mut`, so the test has to look
+            // past it to the bar or comma before *that*. The typed
+            // `|mut tome: u32|` was already caught by its colon, which is what
+            // made the untyped form a quiet false negative rather than a
+            // visible one.
+            let starts_a_param = matches!(before, b'|' | b',')
+                || (prev_words[1] == "mut" && matches!(*prev_word_before, b'|' | b','));
             let untyped_closure_param = in_closure_params
-                && matches!(before, b'|' | b',')
+                && starts_a_param
                 && matches!(rest.as_bytes().first(), Some(b',') | Some(b'|'));
             if (untyped_closure_param || (rest.starts_with(':') && !rest.starts_with("::")))
                 && !*in_where
@@ -6132,6 +6145,7 @@ fn collect_in_code(
                 }
             }
             *prev_byte = bytes[i - 1];
+            *prev_word_before = before;
             if word == "where" {
                 *in_where = true;
             }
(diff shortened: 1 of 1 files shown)


paraphilic-ecchymosis:
Verified: bin suite **46 passed, 0 failed, exit 0** (baseline was 41 before #133 and this), and the four corpus tests from #133 all still pass — which is the check #91 was explicitly waiting for.

The sequencing worked exactly as intended. #91 said the blocker was that no harness existed to re-check the sweep with; #90/#133 built one; this lands against it and the harness confirms no regression on the four shapes the scanner has historically been wrong about. That is a rare case of a dependency between issues being real rather than bureaucratic, and it paid off.

Retargeting to `develop` once #133 merged, rather than leaving it stacked, was the right call — a stacked PR whose base has landed shows a diff nobody can read.

Approving and merging.

Issue #115, opened 2026-09-10
Array element assignment `arr[i] = v` inside a function is silently discarded (but arr[i].f and arr[i][j] propagate)


paraphilic-ecchymosis:
## Summary

Assigning to an array element inside a function — `arr[i] = v` — is **silently
discarded** when the function returns. The write is visible inside the callee
and then vanishes.

Every neighbouring mutation propagates normally:

| mutation inside a function | propagates to caller? |
|---|---|
| `arr[0] = 42` | ❌ **no** |
| `arr[0].field = 42` | ✅ yes |
| `arr[0][0] = 42` | ✅ yes |
| `push(arr, v)` | ✅ yes |
| `s.field = 42` (struct arg) | ✅ yes |

`arr[0][0] = v` working while `arr[0] = v` does not is what makes this look like
an implementation artifact rather than value semantics. There is no error, no
warning, and no diagnostic — the assignment simply does not survive the return.

## Reproduction

```sigil
Σ Holder { n: Int }

rite elem_assign(arr) {
    arr[0] = 42;
    println("  inside fn, arr[0] = " + arr[0]·to_string());
}

rite elem_field_assign(arr) {
    arr[0].n = 42;
}

rite nested_arr_assign(outer) {
    outer[0][0] = 42;
}

rite main() {
    ≔ mut a = [1, 2];
    println("before: a[0] = " + a[0]·to_string());
    elem_assign(a);
    println("after:  a[0] = " + a[0]·to_string());

    ≔ mut objs = [Holder { n: 1 }];
    elem_field_assign(objs);
    println("elem.field: " + objs[0].n·to_string());

    ≔ mut nested = [[1, 2]];
    nested_arr_assign(nested);
    println("nested[0][0]: " + nested[0][0]·to_string());
}
```

```console
$ sigil run ./app.sg
before: a[0] = 1
  inside fn, arr[0] = 42     <-- write is visible here
after:  a[0] = 1             <-- and gone here
elem.field: 42
nested[0][0]: 42
```

Reproduced on `develop` @ `c6ef995`.

## Why it matters

This is a **silent** wrong-answer bug. The callee observes its own write, so
the function looks correct in isolation and under any test that checks state
from inside it. Only the caller sees the loss.

It is also a direct obstacle to decomposition, which is how it was found.
morgoth (`Daemoniorum-LLC/morgoth`) already carries a workaround comment from a
previous session that diagnosed the symptom without identifying the cause:

```sigil
[...]

PR #136, opened 2026-09-10
fix(diagnostics): give ariadne char offsets, not byte offsets — every Sigil diagnostic pointed at the wrong place

### LINKED ISSUE #110: Type errors in a `⊢ Trait ∀ T` method body are reported at the closing brace, not the failing expression
Found while re-measuring the Lares/Qliphoth capability probe against `develop` @ `c6ef995`, compiler built `--no-default-features --features minimal,protocols`.

A type error inside a `⊢ Trait ∀ T` method body is reported at the method's **closing brace** rather than at the failing expression. In a file with several impls the span can land in an entirely unrelated item — in the case that led me here, on a comment and inside an `extern` block a dozen lines below the real call.

## Reproduction

The error is on **line 8**:

```sigil
Θ Go { rite go(self: Self!) -> i64!; }
Σ Wrap { x: i64! }
rite take_bool(b: bool!) -> i64! { 0 }

⊢ Go ∀ Wrap {
    rite go(self: Self!) -> i64! {
        ≔ n = 5;
        take_bool(n)
    }
}

rite main() { println("ok"); }
```

```
[E0003] Error: type mismatch in argument 1: expected bool!, found I64
   ╭─[d-span-impl.sigil:9:6]
   │
 9 │     }
   │      ┬
   │      ╰── type mismatch in argument 1: expected bool!, found I64
```

Note this repro deliberately uses a **struct** target and a plain `i64`/`bool` mismatch, so it shares nothing with #108 — the span is wrong for ordinary type errors in impl bodies.

## The control

The identical mismatch outside an impl is reported correctly, at the call, with the argument underlined:

```sigil
rite take_bool(b: bool!) -> i64! { 0 }

rite main() {
    ≔ n = 5;
    take_bool(n);
}
```

```
 5 │     take_bool(n);
   │                 ┬
   │                 ╰── type mismatch in argument 1: expected bool!, found I64
```

So the span machinery works; something on the impl-body path drops the expression's span and falls back to the enclosing item's end.

Deterministic: 5/5 runs identical (#104 does not apply to `check`).

## Why it is worth fixing rather than living with

With one impl it is a minor annoyance. With several it actively misdirects. Real output from `qliphoth/packages/qliphoth-sys/src/js.sigil` (qliphoth `develop` @ `2ab768b`), six consecutive `⊢ Into[JsValue] ∀ …` impls:

| Error reported at | Actual failing call |
|---|---|
| `342:1` — a `}` | line 323 or 329 |
| `349:1`–`351` — spans two impls | line 335 |
| `357:2` — the comment `// wasm_bindgen` | line 341 |
| `360:29` — `b: bool!` in an `extern` declaration | line 347 |
| `362:41` — `n: i32!` in an `extern` declaration | line 353 |

The last two are the damaging ones: the span points *into the declaration of an unrelated callee*, so the message reads as a defect in the `extern "js"` block. Combined with #108's "expected bool!, found bool!", the file presents as five errors about extern declarations that are character-for-character correct.

The drift is not a constant offset either — it grows with the number of preceding impls — so it cannot be mentally corrected for.

paraphilic-ecchymosis:
Closes #110 — but **not the bug #110 describes**. I filed that issue with the wrong diagnosis, and the real one is considerably larger.

## What I said, and what it actually is

#110 says "type errors in a `⊢ Trait ∀ T` method body are reported at the closing brace". That is a symptom.

`Span` carries **byte** offsets — what the lexer produces and what `&str` indexing uses. `ariadne::Source` indexes by **character**. `write_to` passed byte offsets straight through, so every caret moves forward by one column per extra byte of every multi-byte character before it.

In most languages that is a rounding error. Sigil is written in `⊢ Σ ≔ ⎇ ⎉ ∀ ☉ Θ ·`, three bytes each, so the drift is roughly two columns per symbol and a real file is wrong by whole lines within a screenful.

**Every diagnostic in every Sigil file has been pointing at the wrong place**, and the further into the file the error is, the further off it points. It looked impl-specific only because impl blocks are where the symbols cluster — which is also why the drift appeared to "grow with each preceding impl". Nothing was ever wrong with the spans.

## How it gave itself away

`--format=json` was correct the whole time: `offset_to_line_col` walks `char_indices()` and compares byte indices properly. So the same compiler, from the same span, produced two answers:

```
$ sigil check e.sigil --format=compact     →  line 6, column 19   ← correct
$ sigil check e.sigil                      →  7:2                 ← the closing brace
```

I only found it because I stopped trying to reason about which code path "lost the span" and asked the tool for the span in a second format. Ground truth was already in the tree.

## Measurement

Five shapes, pretty output against `--format=json` as ground truth:

| case | JSON | before | after |
|---|---|---|---|
| plain fn, statement call | 4:15 | 4:17 | **4:15** |
| plain fn, tail call | 4:15 | 4:17 | **4:15** |
| inherent impl | 6:19 | 7:3 | **6:19** |
| trait impl | 7:19 | 8:6 | **7:19** |
| impl, statement call | 6:19 | 7:2 | **6:19** |

The drift in each case equals exactly the extra bytes contributed by multi-byte characters before the error. That arithmetic is what this patch removes.

On the file that prompted the report — `qliphoth/packages/qliphoth-sys/src/js.sigil` — the five errors move from lines **342, 349, 357, 360, 362** (two closing braces, a comment, and twice into an `extern "js"` block that is character-for-character correct) to **329, 335, 341, 347, 353**, which are the calls that actually fail.

Rendered:

```
before                              after
 7 │         0                       6 │         take_bool(n);
   │  ┬                                │                   ┬
   │  ╰── type mismatch …               │                   ╰── type mismatch …
```

## Notes for review

`byte_to_char_offset` is total: it clamps past the end of the source and rounds an offset landing inside a multi-byte character down to that character's start. Rendering a diagnostic is the worst moment to introduce a second failure, so it degrades to a nearby caret rather than panicking.

It is O(n) per offset and called a few times per diagnostic. That is fine for diagnostics and I would rather not cache anything here, but if someone is emitting thousands of errors over a very large file it is the obvious thing to revisit.

| | |
|---|---|
| `cargo test --lib` | **528 passed, 0 failed** (526 + 2 new) |
| Sigil suite | **797 passed, 4 failed** — the four known Kafka/AMQP broker failures |

Counts are under `--features minimal,protocols`, since default features need LLVM 18.

**This changes the rendered position of every diagnostic**, so it is worth a look from anyone with tooling that scrapes the pretty output. Anything reading `--format=json` is unaffected, because that path was already right.

## Related

Found by the Lares capability probe (gap **S126**). It is also another instance of sigil-lang#131 — two outputs answering the same question differently, with the one people read being the wrong one, and nothing in the suite comparing them.

### DIFF
--- a/parser/src/diagnostic.rs
+++ b/parser/src/diagnostic.rs
@@ -189,12 +189,26 @@ impl Diagnostic {
 
     /// Write this diagnostic to a writer.
     pub fn write_to<W: std::io::Write>(&self, writer: W, filename: &str, source: &str) {
-        let span_range: Range<usize> = self.span.start..self.span.end;
+        // `Span` holds BYTE offsets — that is what `&str` indexing and the lexer
+        // produce — while `ariadne::Source` indexes by CHARACTER. Handing it
+        // bytes moves every caret forward by one column per extra byte of every
+        // multi-byte character before it, and the drift accumulates down the
+        // file rather than staying local.
+        //
+        // In most languages that is a rounding error. Sigil is written in
+        // `⊢ Σ ≔ ⎇ ⎉ ∀ ☉ Θ ·`, all three bytes each, so the drift is roughly
+        // two columns per symbol and a real file is wrong by lines within a
+        // screenful. It read as "type errors in impl bodies point at the closing
+        // brace" (#110), because impl blocks are where the symbols cluster —
+        // but the spans were correct all along, and `--format=json`, which uses
+        // `offset_to_line_col` below, always reported them correctly.
+        let to_char = |byte: usize| -> usize { byte_to_char_offset(source, byte) };
+        let span_range: Range<usize> = to_char(self.span.start)..to_char(self.span.end);
 
         let mut colors = ColorGenerator::new();
         let primary_color = self.severity.color();
 
-        let mut builder = Report::build(self.severity.to_report_kind(), filename, self.span.start)
+        let mut builder = Report::build(self.severity.to_report_kind(), filename, span_range.start)
             .with_config(Config::default().with_cross_gap(true))
             .with_message(&self.message);
 
@@ -214,7 +228,7 @@ impl Diagnostic {
         for (span, msg) in &self.labels {
             let color = colors.next();
             builder = builder.with_label(
-                Label::new((filename, span.start..span.end))
+                Label::new((filename, to_char(span.start)..to_char(span.end)))
                     .with_message(msg)
                     .with_color(color),
             );
@@ -274,6 +288,20 @@ impl Diagnostic {
             related: self.related.clone(),
         }
     }
+}
+
+/// Convert a byte offset to a character offset, for consumers that index by
+/// character rather than by byte — `ariadne` being the one that matters here.
+///
+/// Clamps past the end of the source, and rounds an offset that lands inside a
+/// multi-byte character down to that character's start, so a malformed span
+/// degrades to a nearby caret instead of a panic.
+fn byte_to_char_offset(source: &str, byte: usize) -> usize {
+    let mut byte = byte.min(source.len());
+    while byte > 0 && !source.is_char_boundary(byte) {
+        byte -= 1;
+    }
+    source[..byte].chars().count()
 }
 
 /// Convert byte offset to line/column (1-indexed).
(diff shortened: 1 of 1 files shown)


paraphilic-ecchymosis:
Verified, and the failure on `develop` is worse than #110 or this PR describes.

Fixture: 12 struct declarations in Sigil's multi-byte symbols, then an arity error on line 14.

**`develop`:**
```
error: aborting due to 1 previous error
```

That is the entire output. No source excerpt, no caret, no line number — the diagnostic **body is absent**, because the byte offset runs past ariadne's character-indexed source and it renders nothing at all rather than rendering in the wrong place.

**This branch:**
```
 1 │ ╭─▶ Σ P0 { ☉ x: f32 }
14 │ ├─▶ rite main() { ⤺ add(1); }
   │ ╰─────────────────────────────── expected 2 arguments, found 1
```

Lib suite 573 passed / 2 failed (the pre-existing `test_generic_struct_*`).

So the drift does not merely mislocate the caret — past a threshold it silently swallows the whole rendering. Anyone who has hit that saw a bare "aborting due to 1 previous error" and had to go hunting with no location at all. Given today involved reading a great many Sigil diagnostics, I would guess this has cost more time across this ecosystem than any single language bug fixed today, including #115.

**Filing #110 with the wrong diagnosis and then saying so plainly is the right way round.** The symptom genuinely looked impl-specific — and your explanation for why is the convincing part: impl blocks are where the multi-byte symbols cluster, so the drift is worst there, which also explains the reported "grows with each preceding impl". Nothing was ever wrong with the spans. That is a much better account than the issue had.

Merged.

PR #139, opened 2026-09-10
fix(interpreter): resolve `scroll` from the file that declares it

### LINKED ISSUE #138: `scroll` inside a module in a subdirectory cannot name the module beside it
Follow-on from #109, found by walking a morgoth-shaped tree against `develop`
@ `eb85f47` rather than waiting for the report.

#128 made `invoke tome·tui·grid·{Region}` reach a subdirectory. `scroll` did
not follow: it is resolved against the **tome root** wherever it is written, so
a module in a subdirectory cannot name the module beside it.

## Reproduction

```
probe/
├── main.sg
└── tui/
    ├── mod.sg     ☉ scroll vterm;
    └── vterm.sg   ☉ rite vterm_id() -> i32 { 7 }
```

```sigil
// tui/mod.sg
☉ scroll vterm;
☉ rite layer() -> i32 { 1 }

// main.sg
invoke tome·tui·{layer};
☉ rite main() -> !i32 { println(layer()); println(vterm_id()); ⤺ 0; }
```

```console
$ sigil run ./main.sg
1
Runtime error in './main.sg': [R0003] Runtime error: undefined variable: `vterm_id`
```

`layer()` resolves — `tui/mod.sg` loaded — but the `scroll vterm;` inside it
searched `<root>/vterm.sg`, which does not exist. Same for a sibling naming a
sibling: `☉ scroll grid;` inside `tui/vterm.sg` does not find `tui/grid.sg`.

## Why this shape and not another

`scroll` is the spelling morgoth's `lib.sg` already uses for all 18 of its
modules, so it is the one reached for when five of them move into `src/tui/`.
Rust's `mod grid;` in `tui/vterm.rs` names `tui/grid.rs`; `scroll` is the same
declaration and should mean the same thing.

## What *does* work, so the boundary is clear

Checked against `develop` @ `eb85f47`:

| | |
|---|---|
| `invoke tome·tui·grid·{Region}` from anywhere | works (#128) |
| `invoke tome·tui·{layer_name}` → `tui/mod.sg` | works |
| `invoke tome·constants·{esc}` from inside `tui/grid.sg` | works |
| `invoke tome·tui·widgets·button·{…}` (two levels) | works |
| `scroll tui;` at the tome root → `tui/mod.sg` | works |
| **`scroll vterm;` inside `tui/mod.sg`** | **fails** |
| **`scroll grid;` inside `tui/vterm.sg`** | **fails** |

A full morgoth-shaped tree — sibling imports inside a subdirectory, subdirectory
→ top level, top level → subdirectory, types in signatures across the boundary,
impl methods on a subdirectory type, two levels deep — is otherwise correct.

## Cause

`current_source_dir` is the tome root and stays pinned there while a nested
module's items are executed, and the `scroll` handler resolves against it.
`tome·`/`crate·`/`above·` use the same field, and for them that is *right* —
they are rooted at the tome by definition. The two roots are different things
and need to be tracked separately.

## Note on `above·`

`above·constants·{esc}` from `tui/grid.sg` works today, but only because
`above` resolves from the tome root like `tome·` and `constants` happens to be
at the root. Whether `above` should mean "the parent module" is a semantic
question, not a bug — leaving it alone.

Fix in the PR that follows.

paraphilic-ecchymosis:
Closes #138 (does not auto-close — this targets `develop`).

Follow-on from #128. Found by walking a morgoth-shaped tree against `develop`
rather than waiting for the report — morgoth has just gained subdirectories,
and `scroll` is the spelling its `lib.sg` already uses for all 18 of its
modules, so it is the one that gets reached for when five of them move into
`src/tui/`.

## The gap

`scroll` was resolved against the **tome root** wherever it was written, so a
module in a subdirectory could not name the module beside it:

```
probe/
├── main.sg
└── tui/
    ├── mod.sg     ☉ scroll vterm;
    └── vterm.sg   ☉ rite vterm_id() -> i32 { 7 }
```

```console
$ sigil run ./main.sg
1
Runtime error in './main.sg': [R0003] Runtime error: undefined variable: `vterm_id`
```

`tui/mod.sg` loaded — `layer()` resolves — but its `scroll vterm;` searched
`<root>/vterm.sg`. Same for a sibling naming a sibling: `scroll grid;` inside
`tui/vterm.sg` did not find `tui/grid.sg`.

## Two roots, not one

| | rooted at |
|---|---|
| `scroll grid;` | the **declaring file's own directory**, then the tome root |
| `invoke tome·…` / `crate·` / `above·` | the tome root, wherever it is written |

`scroll grid;` in `tui/vterm.sg` names `tui/grid.sg`, the way `mod grid;` in
`tui/vterm.rs` names `tui/grid.rs`. `tome·` is rooted at the tome *by
definition*. Both read `current_source_dir`, which is why they cannot share it
— moving that field around per module would have broken `tome·` from inside a
nested module.

`current_module_dir` carries the declaring file's directory, saved and restored
around each module load alongside `current_module`, at both loaders.

The root stays in the search list **after** the declaring file's directory, so:

- a nested module may still name a top-level one (`scroll constants;` from
  `tui/grid.sg` reaches `constants.sg`) — test included;
- every flat tome is untouched — 798-test suite unchanged;
- when both exist, **the sibling wins** — test included. Searching the root
  first would stop a subdirectory from having a module of its own whenever the
  tome already had one by that name, which is exactly when grouping is wanted.

## The boundary, measured

Checked against `develop` before the change, so the report says what was
actually broken rather than what looked broken:

| | before | after |
|---|---|---|
| `invoke tome·tui·grid·{Region}` from anywhere | ✅ | ✅ |
| `invoke tome·tui·{layer_name}` → `tui/mod.sg` | ✅ | ✅ |
| `invoke tome·constants·{esc}` from inside `tui/grid.sg` | ✅ | ✅ |
| `invoke tome·tui·widgets·button·{…}` (two levels) | ✅ | ✅ |
| `scroll tui;` at the tome root → `tui/mod.sg` | ✅ | ✅ |
| `scroll vterm;` inside `tui/mod.sg` | ❌ | ✅ |
| `scroll grid;` inside `tui/vterm.sg` | ❌ | ✅ |

A full morgoth-shaped tree — `constants.sg`, `render.sg`, `main.sg`, and
`tui/{mod,grid,vterm,layout}.sg` plus `tui/widgets/button.sg`, with a `Region`
built in one subdirectory module, passed through a second, and read in a
third — was already correct and is unchanged.

## `above·`

`above·constants·{esc}` from `tui/grid.sg` works today, but only because
`above` resolves from the tome root like `tome·` and `constants` happens to be
at the root. Whether `above` should mean "the parent module" is a semantic
question, not a bug — left alone deliberately, and called out in the findings
note so it is not mistaken for an oversight.

## Tests

Baseline measured on `develop` (22a43b4) first, same machine, same flags:

Rebased onto `develop` @ `1bdabfc` (was 9 behind; no conflict). Re-measured on
both sides after the rebase:

| | develop @ 1bdabfc | this branch |
|---|---|---|
| `RUST_MIN_STACK=33554432 cargo test --release --no-fail-fast` (lib) | 581 passed / 2 failed | **584 passed / 2 failed** |
| bin | 46 passed / 0 failed | 46 passed / 0 failed |
| `jormungandr/tests/run_tests_rust.sh` | 800 / 4 failed / 2 skipped | 800 / 4 failed / 2 skipped |

`docs/findings/tome-module-paths-and-subdirectories.md` gains the `scroll`
half — the two roots, the search order, why the sibling wins, and the `above·`
note.

### DIFF
--- a/docs/findings/tome-module-paths-and-subdirectories.md
+++ b/docs/findings/tome-module-paths-and-subdirectories.md
@@ -1,6 +1,7 @@
 # How `invoke tome·…` finds a module file
 
-**Recorded:** 2026-09-10, closing #109.
+**Recorded:** 2026-09-10, closing #109; extended the same day for the
+`scroll` half.
 
 A tome used to have to be one flat directory. `invoke tome·tui·grid·{Region}`
 took the module name from the **second** segment alone, looked for `tui.sg`,
@@ -41,6 +42,35 @@ every existing tome uses.
 
 `scroll foo;` goes through the same resolver, so a module may be a file or a
 directory with a `mod.sg` in it.
+
+## `scroll` is relative; `tome·` is not
+
+The two roots differ, and that is the whole reason they are tracked separately:
+
+| | rooted at |
+|---|---|
+| `scroll grid;` | the **declaring file's own directory**, then the tome root |
+| `invoke tome·…` / `crate·` / `above·` | the tome root, wherever it is written |
+
+`scroll grid;` inside `tui/vterm.sg` names `tui/grid.sg`, the way `mod grid;`
+in `tui/vterm.rs` names `tui/grid.rs`. Resolving it against the tome root
+wherever it appeared meant a module in a subdirectory could not name the module
+beside it, and the failure — `undefined variable` — said nothing about why.
+
+The root stays in the search list *after* the declaring file's directory, so a
+nested module may still name one of the tome's top-level modules
+(`scroll constants;` from `tui/grid.sg` reaches `constants.sg`), and every flat
+tome is unaffected. When both exist, the sibling wins: searching the root first
+would stop a subdirectory from having a module of its own whenever the tome
+already had one by that name, which is exactly when grouping is wanted.
+
+`current_module_dir` carries the declaring file's directory and is saved and
+restored around each module load, alongside `current_module`.
+
+`above·` is *not* currently relative — it resolves from the tome root like
+`tome·`. That happens to give the right answer for a top-level module and is
+left alone deliberately; making it mean "the parent module" is a semantic
+change, not a fix.
 
 ## What is still flat
 
--- a/parser/src/interpreter.rs
+++ b/parser/src/interpreter.rs
@@ -1288,6 +1288,15 @@ pub struct Interpreter {
     pub generic_type_bindings: HashMap<String, String>,
     /// Current source directory for resolving relative module paths
     pub current_source_dir: Option<String>,
+    /// The directory of the module file currently being executed, when that is
+    /// not the tome root.
+    ///
+    /// `scroll foo;` is *relative*: inside `tui/vterm.sg` it names
+    /// `tui/grid.sg`, the way `mod grid;` in `tui/vterm.rs` names
+    /// `tui/grid.rs`. `tome·` is not -- it is rooted at the tome wherever it
+    /// is written -- so the two directories are tracked separately rather than
+    /// by moving `current_source_dir` around.
+    pub current_module_dir: Option<String>,
     /// Loaded crates registry (crate_name -> true if loaded)
     pub loaded_crates: HashSet<String>,
     /// Crates currently being loaded (for circular dependency detection)
@@ -1425,6 +1434,7 @@ impl Interpreter {
             current_self_type: None,
             generic_type_bindings: HashMap::new(),
             current_source_dir: None,
+            current_module_dir: None,
             loaded_crates: HashSet::new(),
             loading_crates: HashSet::new(),
             project_root: None,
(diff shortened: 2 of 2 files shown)


paraphilic-ecchymosis:
Verified with the exact shape morgoth is about to write:

```
src/
├── main.sg        invoke tome·tui·{from_sibling};
└── tui/
    ├── mod.sg     scroll grid;  from_sibling() { grid·cell() }
    └── grid.sg    cell() -> 7
```

| | result |
|---|---|
| `develop` | `[R0003] Runtime error` — the sibling cannot be named |
| this branch | **7** |

Lib suite **584 passed**, 2 failed (the pre-existing `test_generic_struct_*`).

**Finding this by walking a morgoth-shaped tree rather than waiting for the report is the part worth calling out.** #128 gave morgoth subdirectories a few hours ago; `scroll` is the spelling its `lib.sg` already uses for all 18 modules, so it is exactly what gets reached for when five of them move into `src/tui/`. You anticipated the consumer instead of waiting to be told, and the gap would otherwise have surfaced as a confusing runtime error in the middle of someone else's refactor.

That is the second time today a fix has exposed the next gap in the same area — #87 and #103 made module-qualified calls resolve, which is what made #115's `Self`-binding bug reachable; #128 made subdirectories work, which is what makes this reachable. Worth expecting a third: each of these enables a shape nobody could previously write, so the next defect is in whatever that shape touches next.

Merged.

PR #140, opened 2026-09-10
fix(interpreter): write an array slot through the array, not a copy of it

### LINKED ISSUE #115: Array element assignment `arr[i] = v` inside a function is silently discarded (but arr[i].f and arr[i][j] propagate)

paraphilic-ecchymosis:
Closes #115 (does not auto-close — this targets `develop`).

> ### ⚠️ One visible behaviour change, called out before you merge
>
> `≔ mut b = a; b[0] = 42;` now leaves `a[0] == 42`. That is the outlier
> joining the other three mutations, **not** aliasing being introduced —
> `push`, `b[0].field = v` and `b[0][0] = v` through the same alias were
> *already* visible through `a` before this change. Evidence below. It is
> pinned by its own test so a reversal fails loudly.
>
> I judged this a one-path fix rather than a semantics decision, for the
> reasons in "Why this is not a semantics change". If you read it the other
> way, this should not land without Daemon.

## Cause

`eval_assign`'s `Expr::Index` arm carried a special case for a **single-segment
path** base — `arr[i] = v`, and nothing else:

```rust
if let Value::Array(arr) = current {
    let borrowed = arr.borrow();
    let mut new_arr = borrowed.clone();          // clone the whole vector
    drop(borrowed);
    if idx < new_arr.len() {
        new_arr[idx] = val.clone();
        self.environment
            .borrow_mut()
            .set(name, Value::Array(Rc::new(RefCell::new(new_arr))))?;  // rebind to a NEW Rc
        return Ok(val);
    }
}
```

It cloned the vector, wrote the element, and repointed the *local binding* at a
new `Rc`. The callee read its own write through the repointed name; the caller
still held the original array. Hence "visible inside, gone outside", with no
error anywhere.

Everything else falls through to the generic path immediately below, which does
`arr.borrow_mut(); borrowed[idx] = val;` — a write through the shared `Rc`.

**The struct-field asymmetry in morgoth #3 is what located it.** `s.items[0] = v`
and `arr[0] = v` both end at "set element 0 of an array", and only the second
was lost — so the difference had to be in how the *base* was reached, not in
what was done to it. A field access is not a single-segment path, so it never
took the special case. Same for `arr[0][0] = v`, whose base is an index.

The fix writes through the `Rc` the binding already holds.

## The table, as tests

One test per row, so the next person can see at a glance which propagate:

| mutation inside a function | before | after | test |
|---|---|---|---|
| `arr[0] = 42` | ❌ | ✅ | `an_array_slot_assigned_in_a_function_reaches_the_caller` |
| `arr[0].field = 42` | ✅ | ✅ | `an_element_field_assigned_in_a_function_reaches_the_caller` |
| `arr[0][0] = 42` | ✅ | ✅ | `a_nested_array_slot_assigned_in_a_function_reaches_the_caller` |
| `push(arr, v)` | ✅ | ✅ | `a_push_in_a_function_reaches_the_caller` |
| `s.items[0] = 42` | ✅ | ✅ | `an_array_reached_through_a_struct_field_reaches_the_caller` |
| `≔ b = a; b[0] = 42` (same scope) | ❌ | ✅ | `an_array_binding_aliases_rather_than_copies` |

Run against `develop` @ `22a43b4` unchanged, **exactly two fail** — the
bare-name slot assignment and the aliasing row. The other four pass on both
sides; they are there to document what already worked, which is the half of the
table that made this bug invisible.

## Why this is not a semantics change

Sigil arrays were already reference-like, and provably so *before* the fix:

```sigil
≔ mut a = [1];  ≔ mut b = a;
push(b, 99);      // len(a) == 2   -- already aliased
b[0].field = 42;  // a[0].field    -- already aliased
b[0][0] = 42;     // a[0][0]       -- already aliased
b[0] = 42;        // a[0] == 1     -- the odd one out
```

The broken path was not value semantics either. A copy would have handed the
caller an independent array; this rebound *one binding* and left every other
alias pointing at the original. It was neither aliasing nor copying — it was a
lost write, which is the one outcome #115 says cannot be right.

Choosing value semantics instead would mean changing `push`, `arr[i].f = v`,
`arr[i][j] = v` and array-through-a-struct-field too, and deciding what
`≔ b = a` means. That would be the language decision; this is not it.

## Verification

Baseline measured on `develop` (22a43b4) first, same machine, same flags:

| | develop | this branch |
|---|---|---|
| `RUST_MIN_STACK=33554432 cargo test --release --no-fail-fast` (lib) | 575 passed / 2 failed | **581 passed / 2 failed** |
| bin | 45 passed / 0 failed | 45 passed / 0 failed |
| `jormungandr/tests/run_tests_rust.sh` | 798 / 4 failed / 2 skipped | 798 / 4 failed / 2 skipped |
| **morgoth `run_tests.sh`** @ `adad4cf` (PR #3) | **239 / 0** | **239 / 0** |

failures are the Kafka/AMQP broker tests.

morgoth is the strongest evidence available: 239 tests over the codebase this
bug actually shaped, run with both binaries via `SIGIL_COMPILER`, in a
throwaway worktree at `adad4cf` — the decompose session's own worktree was not
touched. Nothing anywhere depended on the copy.

## Scope

Interpreter only. The Cranelift JIT rejects a program with an untyped array
parameter outright (`Undefined variable: arr`) both before and after, so it
never reached this shape — a separate pre-existing limitation, not touched here.

`docs/findings/arrays-are-reference-like.md` records the table, the cause, the
asymmetry that located it, the one visible consequence, and what choosing value
semantics would actually require.

### DIFF
--- a/docs/findings/arrays-are-reference-like.md
+++ b/docs/findings/arrays-are-reference-like.md
@@ -1,0 +1,100 @@
+# Sigil arrays are reference-like, and one write pretended otherwise
+
+**Recorded:** 2026-09-10, closing #115.
+
+`arr[i] = v` inside a function was **silently discarded** when the function
+returned. The callee saw its own write; the caller did not.
+
+## The table
+
+| mutation inside a function | propagated before #115 |
+|---|---|
+| `arr[0] = 42` | **no** |
+| `arr[0].field = 42` | yes |
+| `arr[0][0] = 42` | yes |
+| `push(arr, v)` | yes |
+| `s.items[0] = 42` (array through a struct field) | yes |
+
+One row of five. That is what made it dangerous: three neighbouring mutations
+work, so nobody suspects the fourth, and there is no error to suspect it *by*.
+
+## What it cost
+
+It shaped a codebase before anyone identified it. morgoth carried the comment
+
+```sigil
+// Swap focused pane with previous; inline — array slot swap only
+// works in same scope as the panes declaration (Phase 26)
+```
+
+Pane swapping stayed inline in a ~1000-line `main()` because extracting it into
+a `rite` silently stopped working. A compiler bug produced a monofile, and the
+diagnosis stopped at the symptom.
+
+## The cause
+
+`eval_assign`'s `Expr::Index` arm had a special case for a **single-segment
+path** base — `arr[i] = v`, and nothing else:
+
+```rust
+if let Value::Array(arr) = current {
+    let borrowed = arr.borrow();
+    let mut new_arr = borrowed.clone();          // clone the whole vector
+    drop(borrowed);
+    if idx < new_arr.len() {
+        new_arr[idx] = val.clone();
+        self.environment
+            .borrow_mut()
+            .set(name, Value::Array(Rc::new(RefCell::new(new_arr))))?;  // rebind to a NEW Rc
+        return Ok(val);
+    }
+}
+```
+
+It cloned the vector, wrote the element, and repointed the *local binding* at a
+new `Rc`. The callee then read its own write through the repointed name; the
+caller still held the original array.
+
+Every other row falls through to the generic path below it, which does
+`arr.borrow_mut(); borrowed[idx] = val;` — a write through the shared `Rc`.
+`s.items[0]` and `arr[0][0]` never took the special case because their base is
+a field or an index, not a single-segment path. That asymmetry is what located
+the bug: both spellings end at "set element 0 of an array", so the difference
+had to be in how the base was reached.
+
+The fix is to write through the `Rc` the binding already holds.
+
+## Why this is not a semantics change
+
+Sigil arrays were **already** reference-like, and provably so before the fix:
+
+```sigil
+≔ mut a = [1];  ≔ mut b = a;
+push(b, 99);      // len(a) == 2   -- already aliased
+b[0].field = 42;  // a[0].field    -- already aliased
+b[0][0] = 42;     // a[0][0]       -- already aliased
+b[0] = 42;        // a[0] == 1     -- the odd one out
+```
+
+The broken path was not value semantics either. A copy would have given the
+caller an independent array; this rebound one binding and left every *other*
+alias pointing at the original. It was neither aliasing nor copying — it was a
+lost write.
+
+**One visible consequence:** the last line above now leaves `a[0] == 42`. That
+is the outlier joining the other three, not aliasing being introduced. It is
+pinned by `an_array_binding_aliases_rather_than_copies` so a reversal fails
+loudly.
+
+## If you want value semantics instead
+
+That is a language decision, not this fix. It would mean changing `push`,
+`arr[i].f = v`, `arr[i][j] = v` and array-through-a-struct-field as well, and
+deciding what `≔ b = a` means. Silently dropping one write is the only outcome
+that cannot be right, which is what #115 says.
+
+## Scope
+
+Interpreter only. The Cranelift JIT rejects a program with an untyped array
+parameter outright (`Undefined variable: arr`), before and after, so it never
+reached this shape.
--- a/parser/src/interpreter.rs
+++ b/parser/src/interpreter.rs
@@ -5110,7 +5110,18 @@ impl Interpreter {
                     _ => return Err(RuntimeError::new(format!("Index must be an integer, got {:?}", idx_val))),
                 };
 
-                // Get the array and modify it
+                // Get the array and modify it.
+                //
+                // In place, through the `Rc` the binding already holds. This
+                // used to clone the whole vector, set the element, and rebind
+                // the name to a *new* `Rc`, which made `arr[i] = v` the one
+                // mutation that did not survive a return: the callee saw its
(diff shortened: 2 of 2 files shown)


paraphilic-ecchymosis:
Verified independently, and the reasoning is the part I want on the record.

```
rite set_slot(a: [i64]) { a[0] = 42; }
rite main() { ≔ mut arr = [1,2,3]; set_slot(arr); ⤺ arr[0]; }
```

| | result |
|---|---|
| `develop` | **1** — the write is lost |
| this branch | **42** — propagates |

Lib suite **581 passed**, 2 failed — the two pre-existing `llvm_codegen::llvm::tests::test_generic_struct_*`. CI green on all four checks.

**You answered the escalation condition properly rather than routing around it.** I said: if this turns out to change aliasing semantics rather than patch one path, stop and say so, because that would be a language decision. Your answer was that it is not one, and you proved it rather than asserting it — `push`, `arr[0].field` and `arr[0][0]` already aliased *before* the fix, so `arr[0] = v` was the odd one out. And the sharpest line:

> A copy would have handed the caller an independent array; this rebound *one binding* and left every other alias pointing at the original. It was neither aliasing nor copying — it was a **lost write**, which is the one outcome #115 says cannot be right.

Stating what the actual language decision *would* be — changing `push`, `arr[i].f`, `arr[i][j]`, struct-field arrays, and deciding what `≔ b = a` means — is what makes it credible that this is not it.

**The verification matched the stakes.** Running morgoth's own 239 tests with both binaries, in a throwaway worktree at the #3 merge so the decompose session's tree was untouched, plus 798 sigil tests: nothing in either suite depended on the lost write. And scoping honestly — the Cranelift JIT rejects an untyped array parameter outright both before and after, so it never reached this shape, pre-existing and untouched.

`docs/findings/arrays-are-reference-like.md` is the right artefact. This was never written down, which is why three of four neighbouring mutations behaving one way and the fourth behaving another looked like a quirk rather than a bug.

**The consequence worth repeating: this defect shaped an architecture.** morgoth PR #3 established that Phase 26 had to write pane swapping inline because extracting it stopped working — part of why its `main()` reached ~1000 lines. A silent compiler bug produced a monofile, and the author worked around it without ever identifying it. That constraint is now gone.

Approving and merging.

Issue #158, opened 2026-09-10
Hologram·encode is registered twice — a 1-arg stub shadows the real Reed-Solomon encoder


paraphilic-ecchymosis:
Found while fixing #146. `Hologram·encode` is defined **twice** in
`parser/src/stdlib.rs`, and the later registration wins, so the real
implementation is unreachable from Sigil.

## The two registrations

`stdlib.rs:22575` — the real one, 4 arguments, with validation:

```rust
// Hologram·encode(data, data_shards, total_shards, scheme)
define(interp, "Hologram·encode", Some(4), |_, args| {
    let data = args[0].clone();
    let data_shards = …;      // must be a positive integer
    let total_shards = …;     // must be > data_shards
    Ok(create_hologram(data, data_shards, total_shards))
});
```

`stdlib.rs:35821` — a stub, 1 argument, hardcoded parameters:

```rust
define(interp, "Hologram·encode", Some(1), |_, args| {
    let mut fields = HashMap::new();
    fields.insert("__value__".to_string(), args[0].clone());
    fields.insert("__k__".to_string(), Value::Int(3));  // default threshold
    fields.insert("__n__".to_string(), Value::Int(5));  // default total shards
    …
});
```

`define` overwrites, so the stub is what Sigil reaches:

```console
$ sigil run -e 'Hologram·encode("payload", 4, 6, 0)'
Runtime error: Hologram·encode() expects 1 arguments, got 4

$ sigil run -e 'Hologram·encode(42)'
Hologram { __value__: 42, __n__: 5, __k__: 3 }
```

No shards are computed. `create_hologram` and the erasure-coding path behind
it are dead code reachable only from Rust.

## Why this went unnoticed

`jormungandr/tests/stdlib/test_stdlib_Hologram.sg` — headed *"Hologram type
with **REAL assertions**"* — asserts only
`assert_eq(type_of(he), "struct:Hologram")`, which the stub satisfies exactly
as well as the real encoder would. Its other call, `Hologram·new()`, does not
exist at all and was being fabricated by #146's fallback.

## What is needed

Decide which encoder is intended and delete the other. If the 4-arg one is
correct, the stub's callers need updating; if the 1-arg convenience form is
wanted, it should delegate to the real encoder with default shard counts
rather than fabricate fields.

A replacement test must assert something only a real encoder produces — that
the shard count matches `total_shards`, or that the data survives a
reconstruct from a subset. Asserting `type_of` distinguishes nothing.

[...]

Issue #157, opened 2026-09-10
protocol·http·Client has no constructor — P0_008 evidentiality test never built a client


paraphilic-ecchymosis:
Found while fixing #146. `Client·new` is not registered —
`grep '"Client·' parser/src/stdlib.rs` returns nothing — so
`invoke protocol·http·Client; ≔ client = Client·new();` never produced a
client.

## How this stayed invisible

`jormungandr/tests/spec/15_protocols/P0_008_reported_evidence.sg` is
**P0 — Bootstrap Critical**, testing 15-PROTOCOLS.md § 1: that data from an
external protocol carries `~` (reported) evidence and must be explicitly
verified.

```sigil
invoke protocol·http·Client;

async rite main() {
    ≔ client = Client·new();                       // fabricated Client { }
    ≔ response~ = client.get("https://…")|await;
    ≔ verified! = response~|□verify;
    println(1);
}
```

`.expected` is `1`. The only observable output is a literal. Whether the
evidentiality machinery works, or `client` is an empty struct that answers
`get` with nothing, the test prints `1` and passes. It has never tested the
property in its own title.

With #146 the call errors, and the test is marked `// SKIP` referencing this
issue rather than deleted.

## What is needed

A real `Client·new` on the `protocols` feature, and a replacement test that
observes the evidence marker itself — that a response is `~`, that `□verify`
changes it, and ideally that an unverified value cannot be used where `!` is
required. Printing a literal after the pipeline asserts nothing about it.

Note this is the P0 test for the feature; the gap is not that a P0 test fails,
it is that a P0 test has never exercised its subject.

## Related
- #146 — the fallback that hid it
- #131 — the pattern

Issue #156, opened 2026-09-10
QHCompressed has no constructor — the type is unreachable from Sigil


paraphilic-ecchymosis:
Found while fixing #146. `QHCompressed` has **no registered members at all** —
`grep '"QHCompressed·' parser/src/stdlib.rs` returns nothing — so the type
cannot be constructed or used from Sigil under any spelling.

## How this stayed invisible

`jormungandr/tests/stdlib/test_stdlib_QHCompressed.sg` is headed
*"Compressed quantum hypergraph with **REAL assertions**"* and reads:

```sigil
≔ qhc = QHCompressed·new(42);
assert_eq(type_of(qhc), "struct:QHCompressed");
```

#146's fallback fabricated an empty struct **named after the type in the
call**, so `type_of` returned exactly what the assertion expected. The
assertion cannot fail — demonstrated with a control:

```
QHCompressed·new()  ->  struct:QHCompressed
Wibble·new()        ->  struct:Wibble        ← a type invented in that line
```

`assert_eq(type_of(X·new()), "struct:X")` passes for every `X` in existence.

With #146 the call errors, and the test is marked `// SKIP` referencing this
issue rather than deleted, so the gap keeps a number.

## What is needed

Either a constructor and the operations the type is meant to support, or —
if `QHCompressed` is not a type Sigil is meant to expose — its removal from
the stdlib surface and the deletion of the test with it. Right now it is
neither reachable nor absent.

A replacement test must assert a property of the value that could fail:
a round-trip, a compressed size, a recovered element. Not the name it was
constructed under.

## Related
- #146 — the fallback that hid it
- #131 — the pattern

Issue #155, opened 2026-09-10
Rc·downgrade / Weak<T> is unimplemented — 04-MEMORY.md § 4.4 has no runtime


paraphilic-ecchymosis:
Found while fixing #146. `Rc·downgrade` is not registered under any spelling —
`grep '"Rc·' parser/src/stdlib.rs` returns nothing.

`Weak<T>` is a spec feature: **04-MEMORY.md § 4.4 Weak**, and
`jormungandr/tests/spec/04_memory/P1_005_weak_ref.sg` is marked
**P1 — Production Ready**.

## How this stayed invisible

That test has been passing. It could not have been failing, because #146's
fallback fabricated a value for the unimplemented call:

```sigil
≔ rc: Rc<i32> = Rc·new(100);
≔ weak: Weak<i32> = Rc·downgrade(&rc);   // fabricated Rc { }, never used again
println(*rc);                             // 100
≔ rc2 = rc.clone();
println(*rc2);                            // 100
```

`.expected` is `100\n100`. Both lines come from `rc` and `rc2`; **nothing in
the test observes `weak` at all.** It passed identically whether `downgrade`
worked, returned an empty struct, or dropped its argument on the floor.

With #146 the call errors — `no such type 'Rc' (in 'Rc·downgrade')` — which is
correct, and the test is marked `// SKIP` referencing this issue rather than
deleted, so the gap keeps a number someone can find.

## What is actually needed

- `Rc·downgrade(&Rc<T>) -> Weak<T>`
- `Weak·upgrade(&Weak<T>) -> Option<Rc<T>>`, returning `None` after the last
  strong reference is dropped — which is the whole point of the type, and the
  only assertion that can distinguish a real implementation from a stub.

A replacement test must observe the weak reference **after** the strong ones
are gone. Asserting that `downgrade` returned something is the vacuous form,
and is how this got here.

## Related
- #146 — the fallback that hid it
- #131 — the pattern

PR #160, opened 2026-09-10
fix(interpreter): refuse an unknown `Type·member` instead of inventing a value

### LINKED ISSUE #146: An unknown associated function silently evaluates to an empty struct named after the type
`Foo·bar(args)`, where neither `Foo` nor `bar` exists, evaluates to `Foo { }` — an empty struct. The arguments are discarded. No runtime error, and `sigil check` reports no errors.

## Repro

```sigil
☉ rite main() {
    ≔ a = Zzz·qqq(1, 2, 3);
    println("Zzz·qqq(1, 2, 3) = " + a·to_string());
}
```

```
$ sigil check repro.sigil
✓ repro.sigil - no errors
$ sigil run repro.sigil
Zzz·qqq(1, 2, 3) = Zzz {  }
```

Expected a "no such type / no such associated function" error from `check`, or at minimum a runtime error.

## Why this matters beyond typos

It silently defeats calls that *look* implemented. `File·read_to_string` is not registered as an associated function, so it falls through to this fallback:

```sigil
≔ b = File·read_to_string("/etc/hostname");   // a file that exists
println(b·to_string());                        // File {  }
```

It never reads anything, for a file that exists, and reports nothing anywhere. `Fs·read(path)` behaves identically, returning `Fs { }`. Someone who reaches for a `read_to_string` spelling gets an empty struct where they expected file contents, with no signal at any stage — not from `check`, not at runtime.

This is the root cause of the phantom filesystem spellings, and it is the same family as #131: a call reports success while having answered a different question than the reader believes.

---
**Environment:** `develop` @ `aa3d039`, built `cargo build --release --no-default-features --features minimal,protocols`. Found while building `Daemoniorum-LLC/lares-server`, a Sigil HTTP backend probe; standalone repro lives in that repo under `probes/`.

Repro file: `probes/unknown_static_method.sigil`

paraphilic-ecchymosis:
Closes #146 (does not auto-close — this targets `develop`).

> ### Six suite tests go from passing to skipped. **None of them is a regression.**
> They were passing *because* the language fabricated a value for a call to
> something that does not exist. The fix makes the fabrications visible. The
> total is 807 either way; four tests move from the passed column to the
> skipped one, and each carries a filed issue. Please read the table before
> reverting on the count.

## The defect

`eval_path` carried a fallback "for unknown types from external crates". Any
two-segment path whose second segment was spelled like a member returned a
`__constructor__` marker, which `eval_call` turned into an empty struct named
after the type, discarding the arguments. Any capitalised one returned a
made-up enum variant.

```console
$ sigil check repro.sigil
✓ repro.sigil - no errors
$ sigil run repro.sigil
Zzz·qqq(1, 2, 3) = Zzz {  }
```

So `File·read_to_string(path)` returned `File { }` for a file that exists,
having read nothing, and `Fs·read` returned `Fs { }`. Anything shaped like a
call got a plausible value instead of an answer, with no signal at any stage.

## Measured before removing it

Not assumed — instrumented, run per-file with stderr captured:

| corpus | files | firings |
|---|---:|---:|
| `jormungandr/tests` | 896 | **12**, in 11 files |
| morgoth `src/` + `tests/` | 262 | **0** |

Every one of the 12 was a call to something that does not exist under any
spelling: `Superposition·new` where only `·uniform` is registered,
`Rc·downgrade` where Weak is unimplemented, `Client·new` where nothing is.
**Not one was the external-crate type the leniency was written for.** It
existed to furnish defects with values.

> **An earlier draft of this measurement said zero firings, and that was wrong.**
> The runner redirects each test's stderr to a per-test temp file
> (`$TEMP_DIR/${test_name}.err`), so the instrumentation never appeared in the
> runner's own log — which is what I had grepped. Corrected by running each
> file directly and capturing its stderr.
>
> Worth stating why rather than only that, because the next person will reach
> for the log too: a per-file measurement and a log grep look interchangeable
> until you know where stderr goes, and the log is the obvious artefact to
> grep — it is the thing the run hands you. Reading its silence as evidence of
> absence is the same error as reading `git fsck`'s silence as evidence a
> commit is gone: a check that answers a narrower question than the one asked,
> returning nothing, and nothing being mistaken for an answer.
>
> The corrected number is also the more useful one. Zero firings would have
> suggested the fallback was inert and the fix cosmetic; twelve firings, every
> one masking a missing implementation, is what makes the case.

## What the six tests were doing

Four stdlib files are headed *"…with **REAL assertions**"* and assert only:

```sigil
≔ sp = Superposition·new();
assert_eq(type_of(sp), "struct:Superposition");
```

The fallback named the fabricated struct **after the type in the call**, so
this passes for every possible `X` — demonstrated with a control:

```
Superposition·new()  ->  struct:Superposition
Wibble·new()         ->  struct:Wibble        ← a type invented in that line
```

`P1_005_weak_ref` is the sharpest. It is **P1 — Production Ready** for
04-MEMORY.md § 4.4, and it calls the unimplemented `Rc·downgrade`, **discards
the result**, and passes on `100` / `100` printed by two unrelated lines.
Nothing in it observes a weak reference.

## Disposition

**Skipped — implementation absent, issue filed, test kept not deleted:**

| test | called | reality | issue |
|---|---|---|---|
| `P1_005_weak_ref` | `Rc·downgrade` | Weak unimplemented | #155 |
| `test_stdlib_QHCompressed` | `QHCompressed·new` | no members registered at all | #156 |
| `P0_008_reported_evidence` | `Client·new` | no constructor | #157 |
| `test_stdlib_Hologram` | `Hologram·new` | a 1-arg stub **shadows** the real 4-arg encoder | #158 |

#158 was found along the way and is its own defect: `Hologram·encode` is
registered twice and the later stub wins, so the real Reed-Solomon encoder is
unreachable from Sigil. That is why Hologram is skipped rather than re-pointed
— its reachable constructor is itself a stub, and a test asserting the stub's
behaviour would be the same mistake one level down.

**Re-pointed to the constructor that exists, with assertions that can fail:**

| test | now uses | asserts |
|---|---|---|
| `test_stdlib_Superposition` | `Superposition·uniform` | elements preserved; one amplitude per element; amplitudes **normalised** (squares sum to 1); a 2-element state has larger amplitudes than a 3-element one |
| `test_stdlib_MerkleTree` | `MerkleTree·from` | root is **deterministic** (same leaves → same root) **and binding** (different leaves → different root); root is a 64-char digest; a changed element changes only its own leaf hash |

**Verified by mutation, not by inspection.** I patched a constant Merkle root
and un-normalised amplitudes into the stdlib and re-ran:

```
MUTANT build:
  test_stdlib_Superposition  ->  Assertion failed
  test_stdlib_MerkleTree     ->  Assertion failed
  old-style type_of asserts  ->  PASSED          ← on the same mutant
```

The new assertions catch what the old ones could not. That check exists
because replacements of this kind are themselves easy to write vacuously.

## The suite now reports skips honestly

`TOTAL` was `PASS + FAIL`, so skips vanished from the pass rate and
"797 passed, 6 skipped" printed the same 100% as "797 passed, 0 skipped".
Now:

```
✅ Passed: 797
❌ Failed: 4
⏭  Skipped: 6
Pass rate: 98% (797 of 807; skips count as not passing)

6 test(s) skipped because the implementation is ABSENT, not because
they are flaky. Each names its issue above. A skip is a missing feature
that is being counted rather than hidden; it should shrink, not settle.
```

Each skip line echoes its issue number, so a reader of the run reaches the gap
without opening the file:

```
⏭  SKIP: P1_005_weak_ref -- #155 -- Rc·downgrade / Weak<T> is UNIMPLEMENTED (04-MEMORY.md § 4.4).
```

The wording is deliberate: *absent, not flaky*. Those two get conflated within
a month, and then the skip becomes permanent furniture.

## Tests

Measured on `develop` (2a8f1a9) first, same machine, same flags:

| | develop | this branch |
|---|---|---|
| `RUST_MIN_STACK=33554432 cargo test --release --no-fail-fast` (lib) | 586 passed / 2 failed | **590 passed / 2 failed** |
| `jormungandr/tests/run_tests_rust.sh` | 801 / 4 failed / 2 skipped | 797 / 4 failed / **6 skipped** |

`llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}`; the 4
suite failures are the Kafka/AMQP broker tests.

Four regression tests added, covering both arms (`Zzz·qqq` and `Zzz·Qqq`), the
known-type-unknown-member case, and that a real enum still answers for its own
variants while a bad variant on it still says so.

## Scope

Interpreter only, one hunk in `eval_path`. #124 (`Sys·socket` simulating) is a
**separate** defect and this does not touch it — proven with one program:
`Sys·socket(2,1,0)` → `21000` (registered, executes, simulates) versus
`Sys·nosuchcall(1)` → `Sys { }` (unregistered, this fallback). Neither fix
closes the other.

### DIFF
--- a/jormungandr/tests/run_tests_rust.sh
+++ b/jormungandr/tests/run_tests_rust.sh
@@ -100,9 +100,17 @@ run_test() {
     local test_out="$TEMP_DIR/${test_name}.out"
     local test_err="$TEMP_DIR/${test_name}.err"
 
-    # Check if test should be skipped
+    # Check if test should be skipped.
+    #
+    # A `// SKIP` marks a test whose implementation is ABSENT -- the feature
+    # does not exist, so the test cannot pass. It is not a flaky test and not
+    # a disabled one. The marker carries the issue number for the gap, and it
+    # is echoed here so a reader of the run reaches the gap without opening
+    # the file.
     if grep -q "^// SKIP" "$test_file" 2>/dev/null; then
-        echo -e "  ${YELLOW}⏭  SKIP${NC}: $test_name (marked as SKIP)"
+        local skip_reason
+        skip_reason=$(grep -m1 "^// SKIP" "$test_file" | sed 's|^// SKIP *||')
+        echo -e "  ${YELLOW}⏭  SKIP${NC}: $test_name -- ${skip_reason}"
         ((SKIP++))
         return 0
     fi
@@ -250,10 +258,21 @@ if [ $TIMEOUT -gt 0 ]; then
 fi
 echo -e "${BLUE}───────────────────────────────────────────────${NC}"
 
-TOTAL=$((PASS + FAIL))
+# Skips count against the total. Excluding them let "795 passed, 6 skipped"
+# and "795 passed, 0 skipped" print the same 100%, which is the one thing the
+# number must never do -- a skipped test is a feature that does not exist, and
+# hiding it makes the suite report a capability the compiler does not have.
+TOTAL=$((PASS + FAIL + SKIP))
 if [ $TOTAL -gt 0 ]; then
     PERCENTAGE=$((PASS * 100 / TOTAL))
-    echo -e "${BLUE}Pass rate: ${PERCENTAGE}%${NC}"
+    echo -e "${BLUE}Pass rate: ${PERCENTAGE}% (${PASS} of ${TOTAL}; skips count as not passing)${NC}"
+fi
+
+if [ $SKIP -gt 0 ]; then
+    echo
+    echo -e "${YELLOW}${SKIP} test(s) skipped because the implementation is ABSENT, not because${NC}"
+    echo -e "${YELLOW}they are flaky. Each names its issue above. A skip is a missing feature${NC}"
+    echo -e "${YELLOW}that is being counted rather than hidden; it should shrink, not settle.${NC}"
 fi
 
 # Exit with error if any tests failed
--- a/jormungandr/tests/spec/04_memory/P1_005_weak_ref.sg
+++ b/jormungandr/tests/spec/04_memory/P1_005_weak_ref.sg
@@ -1,3 +1,14 @@
+// SKIP #155 -- Rc·downgrade / Weak<T> is UNIMPLEMENTED (04-MEMORY.md § 4.4).
+//
+// The implementation is ABSENT, not flaky. This test does not intermittently
+// fail; it cannot pass until #155 lands.
+//
+// It passed until #146 only because the fallback fabricated a value for the
+// unimplemented `Rc·downgrade`, whose result this test never reads -- the
+// expected `100\n100` comes from `rc` and `rc2` on unrelated lines. It
+// asserted nothing about weak references.
+//
+// Kept rather than deleted so the gap keeps a number a reader can find.
 // Test: Weak references with Rc
 // Spec: 04-MEMORY.md § 4.4 Weak
 // Priority: P1 - Production Ready
--- a/jormungandr/tests/spec/15_protocols/P0_008_reported_evidence.sg
+++ b/jormungandr/tests/spec/15_protocols/P0_008_reported_evidence.sg
@@ -1,3 +1,14 @@
+// SKIP #157 -- `protocol·http·Client` has no constructor; `Client·new` is
+// not registered under any spelling.
+//
+// The implementation is ABSENT, not flaky.
+//
+// It passed until #146 only because the fallback fabricated a `Client { }`.
+// The expected output is the literal `1`, printed after the pipeline, so the
+// test never observed the evidentiality it is named for -- a P0 test that has
+// not exercised its subject.
+//
+// Kept rather than deleted so the gap keeps a number a reader can find.
 // Test: Protocol data is always reported (~)
 // Spec: 15-PROTOCOLS.md § 1 Philosophy
 // Priority: P0 - Bootstrap Critical
--- a/jormungandr/tests/stdlib/test_stdlib_Hologram.sg
+++ b/jormungandr/tests/stdlib/test_stdlib_Hologram.sg
@@ -1,3 +1,13 @@
+// SKIP #158 -- the reachable `Hologram·encode` is a 1-arg stub that shadows
+// the real 4-arg Reed-Solomon encoder; `Hologram·new` does not exist.
+//
+// The implementation is SHADOWED, not flaky.
+//
+// `Hologram·new` was fabricated by the fallback removed in #146, and the
+// `type_of` assertion on `encode` is satisfied equally by the stub and by a
+// real encoder, so this file could not tell them apart.
+//
+// Kept rather than deleted so the gap keeps a number a reader can find.
 // Stdlib Tests: Hologram
 // Hologram type with REAL assertions
 
--- a/jormungandr/tests/stdlib/test_stdlib_MerkleTree.sg
+++ b/jormungandr/tests/stdlib/test_stdlib_MerkleTree.sg
@@ -1,16 +1,38 @@
 // Stdlib Tests: MerkleTree
-// Merkle tree data structure with REAL assertions
+//
+// Constructed through `MerkleTree·from`, which is the constructor that exists.
+// This file previously called `MerkleTree·new`, which does not, and asserted
+// `type_of(mt) == "struct:MerkleTree"` -- an assertion that could not fail,
+// because the fallback removed in #146 named the struct it fabricated after
+// the type in the call.
+//
+// The properties below are the ones that make a Merkle tree a Merkle tree:
+// the root is determined by the contents, and it distinguishes them.
 
 rite main() {
     println("=== Testing MerkleTree ===");
 
-    // MerkleTree·new - creates empty merkle tree
-    ≔ mt = MerkleTree·new();
-    assert_eq(type_of(mt), "struct:MerkleTree");
+    ≔ a = MerkleTree·from(["a", "b"]);
+    ≔ b = MerkleTree·from(["a", "b"]);
+    ≔ c = MerkleTree·from(["a", "c"]);
 
-    // MerkleTree·from - creates tree from data
-    ≔ mtf = MerkleTree·from(["a", "b", "c", "d"]);
-    assert_eq(type_of(mtf), "struct:MerkleTree");
+    // Deterministic: the same leaves give the same root. A root derived from
+    // anything but the contents -- a counter, an address, a random value --
+    // fails here.
+    assert_eq(a.root, b.root);
+
+    // Binding: different leaves give a different root. A constant root, or one
+    // that ignores its input, passes the line above and fails this one. The
+    // two together are what a stub cannot satisfy.
+    assert(a.root != c.root);
+
+    // A root is a SHA-256 hex digest, so 64 characters, and not empty.
+    assert_eq(len(a.root), 64);
+
+    // One leaf hash per element, and a changed element changes its own leaf.
+    assert_eq(len(a.__leaf_hashes__), 2);
+    assert_eq(a.__leaf_hashes__[0], c.__leaf_hashes__[0]);
+    assert(a.__leaf_hashes__[1] != c.__leaf_hashes__[1]);
 
     println("ALL MERKLETREE TESTS PASSED!");
 }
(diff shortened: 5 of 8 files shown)


paraphilic-ecchymosis:
Ran `lares-server` against this branch before it merges. **It does what it says, and it does not disturb anything else here.**

| toolchain | commit | features |
|---|---|---|
| control | `develop` @ `2a8f1a9` | `--no-default-features --features minimal,protocols` |
| under test | `fix/146-unknown-associated-fn` @ `97556f1` | same |

Both from throwaway detached worktrees. `develop` had moved since my earlier measurements at `aa3d039`, so the control was rebuilt rather than reused — otherwise "#160 changed this" and "develop drifted" are the same observation.

## 5 of 15 probes change verdict, and all five are #146

| probe | develop `2a8f1a9` | #160 `97556f1` |
|---|---|---|
| `unknown_static_method` | exit 0, `Zzz { }` | **exit 1** — `no such type 'Zzz' (in 'Zzz·qqq')` |
| `unknown_method_on_declared_type` | exit 0, `Real { }` | **exit 1** — `no associated function or variant 'nope' on 'Real'` |
| `unknown_static_discards_args` | exit 0, both args evaluated | **exit 1**, before any argument is evaluated |
| `unknown_static_case_asymmetry` | exit 1, after fabricating | **exit 1**, refuses the lowercase spelling too |
| `check_misses_two_segment_paths` | exit 0, `Zzz { }` | **exit 1** |

Splitting *"no such type"* from *"no member on a type that exists"* is the half the empty struct never told anyone, and it lands correctly on the declared-struct case.

**`unknown_static_discards_args` settles the scoping question empirically.** On `develop` both arguments are evaluated — side effects printed — and then discarded. On this branch the refusal happens *before either runs*. One fix, both symptoms, as scoped.

## The other ten are byte-identical, which is the more useful result

Same output, same exit codes, both toolchains: `block_then_minus` (#121), `no_concurrency` (#122), `sys_socket_is_simulated` (#124), `string_to_number` (#125), `json_parse_aborts` (#147), `fs_is_not_total` (#148), `no_fsync` (#149), `char_vs_string_eq` (#150), `vec_aliases_not_moves` (#151), `pipe_unresolved_names` (#159).

So the six gaps filed off this probe are **six defects, not one root cause in different clothes.** #146 was the widest and it closes without touching any of the others.

`pipe_unresolved_names` being unchanged also confirms empirically what #159 argued from the source: the pipe gap and #146 share only the checker. This branch changes pipe behaviour not at all.

## No collateral damage to a real program

- `scripts/verify.sh` — **23/23 on both.** I had pre-declared that this count might drop once strict behaviour arrived; **it did not, and I was wrong about why it would.** The fallback this removes sits *after* the globals lookup, so the 25 registered-builtin call sites in `src/` (`Vec·new`, `HashMap·new`, `TcpStream·read`, `Sys·getenv`, `String·from_utf8`, …) never reach it.
- `sigil check src/*.sigil` — 4/4 clean on both.

The server binds a real socket, serves 15 endpoints' worth of checks, restarts with its store intact and survives a corrupted store file, identically on both builds.

## One thing worth knowing, not a blocker for this PR

`sigil check --strict` flags **zero** qualified paths — but on **both** toolchains it fails `src/main.sigil` (15 names) and `src/store.sigil` (5), all of which exist: `append`, `load`, `save`, `parse_request`, `json_response` … reached through `invoke store·*` / `invoke http·*` / `invoke json·*`, plus free-function builtins `fs_read`, `fs_write`, `fs_rename`, `fs_exists`.

Pre-existing and unrelated to this change, but it makes `--strict` unusable on any multi-module program today. Belongs on #159; noting it here only so it is not mistaken for fallout from this branch when someone next runs `--strict`.

Repros are in `Daemoniorum-LLC/lares-server` under `probes/` (PR #2, pending).

paraphilic-ecchymosis:
Merged. CI green on all four jobs (ubuntu + macos Build and Test, MCP Server, VSCode Extension), and independently verified by lr-adhoc-17 against a compiler built from a clean `origin/develop` worktree.

This is the highest-blast-radius fix in the ecosystem. Before it, `Zzz·qqq(1,2,3)` on an undefined type with an undefined method evaluated to `Zzz {}` with the arguments silently discarded, which is why `File·read_to_string("/etc/hostname")` returned `File {}` for a file that exists. The language was fabricating a plausible value for code naming nothing real.

## Two things the verification established that the description does not

**`sigil check` catches none of this, before or after.** Exit 0 and "no errors" on the `Zzz·qqq` repro under both the old and new builds — this is a **runtime-only** refusal. So a check-clean count remains an upper bound on what is verified, not a floor, and nobody should read this merge as making `check` trustworthy. That is sigil-lang#118 and it is untouched here.

**The six newly-failing spec tests are fabrications becoming visible, not a regression.** Three were re-pointed at real constructors; three call things that genuinely do not exist under any spelling — `QHCompressed·new`, `Client·new`, `Rc·downgrade` — and are skipped with an issue filed each. `P1_005_weak_ref` is the sharpest: it called the unimplemented `Rc·downgrade`, discarded the result, and passed on `100/100` printed by unrelated lines. A pass count falling because the tools stopped fabricating is the measurement improving.

## A correction to my own instructions, which affected this PR

I have been telling sessions to build with `--no-default-features --features minimal,protocols,react-migrate,wasm`. **Those flags produce a false green.** On one commit:

```
minimal flags : 920 passed / 0 failed     ← 100% green
default       : 586 passed / 2 failed
```

The sets are **disjoint** — `minimal` drops `llvm_codegen` entirely (47 tests, including both known reds) and adds `migrate::react::*` and `wasm::*`. Any count from a minimal build is not comparable to a figure in a PR body, and it hides the two failures the PRs correctly disclose. I put that flag set in two verification briefs and it was found propagating verbatim into a third session's prompt. Corrected everywhere; re-measure or mark as not comparable.

## For #163

These two conflict **textually but not semantically** — one hunk, both appending tests at the same point in `mod tests`; the production hunks (`eval_path` vs `process_use_tree`) never overlap. They compose: 592 P / 2 F lib, 797/4/6 suite, all four new tests passing. The resolution is load-bearing rather than cosmetic — two attempts failed to compile on brace balance, because the conflict begins after a shared `#[test]` line that must be re-added. Rebase onto this and resolve by keeping both blocks.

paraphilic-ecchymosis:
Follow-up: **re-measured under default features, and everything in my earlier comment holds.**

My original numbers came from `--no-default-features --features minimal,protocols`, on the strength of a brief saying default features need LLVM 18 and it is not installed here. That premise is false on this machine — `llvm-config-18` reports 18.1.3 at `/usr/lib/llvm-18` and a default `cargo build --release` succeeds. Re-run with both feature sets at the same commit (`develop` @ `652dc58`, which now contains this PR): all 15 probes byte-identical, `verify.sh` 23/23 on both, `check`/`--strict` identical. Nothing here depends on the feature set.

One correction to my earlier comment worth flagging, because it nearly became a wrong claim about this PR. **This merged at 21:43:30 UTC, between my two `git worktree add` calls.** So my first minimal-vs-default comparison had the default build at `2a8f1a9` (pre-merge) and the minimal build at `652dc58` (post-merge) — two variables changed at once, and the five differing probes looked like a feature-flag effect when they were in fact this PR working exactly as intended.

Caught by `rev-parse`-ing each worktree rather than trusting the line I had printed asserting both were at the same commit. The lesson for anyone else measuring against a branch today: **pin the commit SHA in `worktree add`, never name a ref** — `develop` is moving fast enough that two adds minutes apart can straddle a merge.

The verdicts in my earlier comment were all taken against `97556f1` explicitly and are unaffected.

paraphilic-ecchymosis:
**Correction to my merge comment above.** I wrote "CI green on all four jobs" and offered it as evidence. It was not evidence, and the merge did not actually rest on it.

`.github/workflows/ci.yml` carries `continue-on-error: true` on **four** steps — Run tests (line 65), Check formatting (70), Run clippy (75), and the examples check (83). The raw logs show `Process completed with exit code 101` on both the ubuntu and macOS runs, swallowed. The job reports success by construction. There is also no `--no-fail-fast`, so the run halts at the first failing binary and `main.rs`'s tests never execute.

So the green tick answered "did the job step complete", not "did the tests pass" — while I quoted it in the vocabulary of the second. Filed as #175.

**What this merge does rest on**, unchanged and still sound: lr-adhoc-17's independent verification against a compiler built from a clean `origin/develop` worktree, including the `Zzz·qqq` repro reproduced under both the old and new builds, and the finding that `sigil check` catches none of it because the refusal is runtime-only. None of that came from CI.

I am correcting only this comment. I checked the other four sigil-lang PRs I merged today — #154, #163, #168, #169 — and none of them cites CI as evidence; each records the local verification instead. The exposure is one sentence in one comment, and I would rather state that precisely than apologise more broadly than the facts support.

Worth naming what this is: I have spent the day asking every session what question their green signal actually answered, and then quoted one without asking it myself. The check that caught me was somebody else reading a workflow file I never opened.

PR #163, opened 2026-09-10
fix(interpreter): reach a sibling module from an `invoke` inside a subdirectory

### LINKED ISSUE #152: Subdirectory module cannot bind a constant from a sibling: works flat, fails nested, masked when the caller imports it too
## Summary

A module in a **subdirectory** cannot bind a top-level constant from a sibling
module. `invoke tome·consts·{MARK}` inside `lib/user.sg` does not make `MARK`
available in `user.sg`'s own body — it fails at the use site with
`undefined variable: MARK`.

The same code one directory up works. And the failure is **masked** whenever the
top-level program happens to import the same constant, which is what makes it
dangerous: the module appears to work, and only breaks for a consumer that does
not already import its dependencies.

Types are unaffected — `invoke tome·grid·{Grid, Region}` from a subdirectory
module resolves fine. This is specific to constants (values).

## Reproduction

**Works — flat.**

```
flat/
├── app.sg      invoke tome·user·{emit};        rite main() { println(emit()); }
├── user.sg     invoke tome·consts·{MARK};      rite emit() { ↩ "[" + MARK + "]"; }
└── consts.sg   ≔ MARK = "M";
```

```console
$ sigil run ./app.sg
[M]
```

**Fails — the identical modules in a subdirectory.**

```
nested/
├── app.sg          invoke tome·lib·user·{emit};   rite main() { println(emit()); }
└── lib/
    ├── user.sg     invoke tome·consts·{MARK};     rite emit() { ↩ "[" + MARK + "]"; }
    └── consts.sg   ≔ MARK = "M";
```

```console
$ sigil run ./app.sg
Runtime error in './app.sg': [R0003] Runtime error: undefined variable: `MARK`
```

`lib/user.sg`'s sibling import is byte-identical in both cases.

**Masked — add the constant to the caller and the same nested tree works:**

```sigil
invoke tome·lib·user·{emit};
invoke tome·lib·consts·{MARK};     // caller does not use MARK itself

rite main() { println(emit()); }   // -> [M]
```

Measured on `develop` @ `aa3d039`.

## Why the masking matters

This is what it looks like in practice. morgoth just moved its terminal layer
into `src/tui/`. Three of those modules import shared constants from a sibling:

```sigil
// src/tui/terminal.sg
invoke tome·constants·{ESC};        // 12 uses — alt screen, mouse, cursor
```

`terminal_init()` **fails** when called from a program that imports only
`tome·tui·terminal`, and **works** inside morgoth — because `src/render.sg` and
`src/monitor.sg` independently import `tome·tui·constants`, putting `ESC` in
scope for everything.

So:

* the whole suite is green (239/239),
* every module type-checks,
* the application renders correctly,
* and the layer is **not** usable by anything else.

The bug is invisible from inside the application that has the accident, and
only appears the moment someone tries to reuse the module — which is exactly
the case a subdirectory namespace exists to support.

`sigil check` does not catch it either: an unresolved `invoke` is not an error
in itself, it fails at the use site.

## Suggested test

A regression test wants the *unmasked* shape specifically — a nested module
whose constant dependency the top-level program does **not** import. A test
where the caller also imports it passes either way.

---
Found extracting morgoth's terminal layer into `src/tui/` (follows #128).

paraphilic-ecchymosis:
## Summary

`invoke tome·consts·{MARK}` inside `lib/user.sg` did not bind `MARK` for
`user.sg`. The identical modules one directory up worked.

`scroll` was taught to search the declaring file's own directory before the
tome root in #139. `invoke` was not — it resolved only against
`current_source_dir`, the tome root — so a module in a subdirectory could not
name the module beside it. This gives the `invoke` path the same search list,
declaring directory first, root second, for the reasons already recorded on the
`scroll` side: an `invoke` is relative, and the root stays in the list because
every flat tome depends on it.

One hunk in `process_use_tree`. Does not touch `eval_path`, so it does not
collide with #146.

## Why this went unnoticed, which is the interesting part

Two things hid it, and they compound.

**The failure is silent.** Not finding a module file returns `Ok(())`. Nothing
binds, nothing errors, `sigil check` passes. The first symptom is `undefined
variable` at a use site — which reads as a bug in the module, not in the loader.

**It disappears whenever the top-level program imports the same name.** That
puts the name in scope for everything, including the nested module whose own
binding never worked. The module then *works by coincidence*, and breaks for the
next consumer that does not happen to import its dependencies.

morgoth shipped in exactly that state. `src/tui/terminal.sg` does
`invoke tome·constants·{ESC}`; `terminal_init()` worked — but only because
`src/render.sg` independently imported the same constants. Its suite was green,
every module compiled, the application rendered correctly, **and the layer could
not be used by anything else.**

## Tests

`a_nested_module_binds_a_constant_from_its_sibling` is the regression test, and
its consumer **deliberately does not import the constant**. That shape is the
whole point — the masked form passes with or without this fix.

`the_masking_import_is_not_a_regression_test_for_that` pins that down: same
tome, consumer also imports the constant, passes either way. Kept so nobody
writes only that shape and concludes the bug is gone.

Constants are the gap rather than modules — the neighbouring nested-module
tests all cover `☉ rite`, and functions were already reached.

## Verified

**Rebased onto `develop` @ `652dc58`** (after #160 merged). The conflict was
textual only — both PRs append tests at the same point in `mod tests`, while
the production hunks never meet: mine is in `process_use_tree`, #160's is in
`eval_path`. Resolved keeping both sides; it needs the `#[test]` re-added
between them, because that attribute was shared *above* the marker and the
closing brace shared *below* it, so each side carries a brace delta of +1.

**Base:** `develop` @ `652dc58` (this branch: `c166bc3`).
**Features:** default (`jit, llvm, protocols, native`).
**Command:** `RUST_MIN_STACK=33554432 cargo test --release --no-fail-fast`

| binary | passed | failed |
|---|---|---|
| lib (`src/lib.rs`) | 592 | 2 |
| bin (`src/main.rs`) | 49 | 0 |
| **all binaries** | **641** | **2** (643 total) |

The two reds are the pre-existing
`llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}`. All six new
tests pass — #160's four and this PR's two.

`--no-fail-fast` matters and so does dropping `--lib`: cargo stops after the
first failing binary otherwise, and a lib-only figure silently omits the 49
tests in the bin target. An earlier revision of this note quoted 592/594, which
is the lib binary alone — correct for what it measured, and not the total.

For cross-checking against figures measured on `develop` @ `2a8f1a9`
(635 passed / 2 failed, 637 total): this branch's 643 is that plus #160's four
tests plus this PR's two, all six of which pass.

The unmasked test fails before this change and passes after; the masked one
passes both times, as documented.

### On feature flags and comparability

An earlier revision of this PR said "543/543" with no base or flags. Measured on
this commit by listing test names under each set:

| build | tests | vs default: dropped | vs default: added |
|---|---|---|---|
| default | 594 | — | — |
| `minimal,protocols` *(what produced the 543)* | 547 | **47** (all `llvm_codegen`) | **0** |
| `minimal,protocols,react-migrate,wasm` | 926 | 47 | **379** (231 `wasm`, 148 `migrate`) |

The two sets **overlap heavily and neither contains the other** — 547 tests are
common to both on this commit:

|  | default | four-flag |
|---|---|---|
| shared | 547 | 547 |
| only in this build | 47 (`llvm_codegen`) | 379 (231 `wasm`, 148 `migrate`) |
| **total** | **594** | **926** |

Two different problems, worth not conflating:

- `minimal,protocols` is a **strict subset** of default: the same 47 dropped, 0
  added. Every test in it is a real test, but both known reds are among the
  missing 47 — so `543/543` was all-green *because* the failures had not been
  built. A false green, and the number said nothing about it.
- `minimal,protocols,react-migrate,wasm` is **not a subset in either
  direction**: it omits the same 47 (both reds with them) and adds 379 that a
  default build never compiles. Its totals are not comparable to a default
  total either way, and a clean 920/0 from it still hides the two failures the
  PRs disclose.

An earlier revision of this note called that second case "disjoint". That was
wrong — the sets share 547 tests. The load-bearing property is that neither is
a superset of the other, which is a weaker and more precise claim, and the one
that actually makes the counts incomparable.

The remedy differs — a subset under-reports; a disjoint set reports something
else — which is why the distinction is here rather than a single "minimal flags
are wrong" note.

Against morgoth, whose reusable-layer finding this unblocks:

| | before | after |
|---|---|---|
| `terminal_init()` via `tome·tui·terminal`, caller imports no constants | `undefined variable: ESC` | OK |
| `render_border()` via `tome·tui·draw`, same | `undefined variable: BOX_H` | `┌` |

The full `src/tui/` layer copied into an empty directory **with no morgoth
module present** now runs a non-morgoth program end to end — layout, border,
`fill_row`, vterm feed + blit, text overlay. morgoth's own suite: **239/239,
exit 0, twice** (morgoth's runner prints its own compiler provenance per run).

## Notes

Based on `develop` @ `2a8f1a9`. #146 was not yet merged when this branched, so
expect a trivial add/add in the test region around the nested-module tests —
coordinated with that session in advance; the fixes themselves are in different
functions.

Closes #152 (needs closing by hand).

### DIFF
--- a/parser/src/interpreter.rs
+++ b/parser/src/interpreter.rs
@@ -4159,8 +4159,29 @@ impl Interpreter {
                         if !self.types.contains_key(&qualified)
                             && self.globals.borrow().get(&qualified).is_none()
                         {
-                            // Load from current source directory
-                            if let Some(source_dir) = &self.current_source_dir.clone() {
+                            // Searched from the declaring file's own directory
+                            // first, then the tome root -- the same list a
+                            // `scroll` uses (#139). An `invoke` is relative for
+                            // the same reason: `invoke tome·constants·{ESC}`
+                            // inside `tui/terminal.sg` means `tui/constants.sg`.
+                            //
+                            // Only the root was searched before, so a module in
+                            // a subdirectory could not reach its own sibling.
+                            // It failed silently -- the `else` below returns
+                            // Ok(()) -- so nothing bound, `sigil check` passed,
+                            // and the first use site said `undefined variable`.
+                            // Worse, it was invisible whenever the top-level
+                            // program imported the same name, which put it in
+                            // scope for everything: the nested module then
+                            // worked by coincidence, and broke for the next
+                            // consumer that did not import it.
+                            let search_dirs: Vec<String> = self
+                                .current_module_dir
+                                .iter()
+                                .chain(self.current_source_dir.iter())
+                                .cloned()
+                                .collect();
+                            if !search_dirs.is_empty() {
                                 // Which of the path's segments name the module,
                                 // and which name the item inside it, is not
                                 // known from the path alone:
@@ -4178,8 +4199,9 @@ impl Interpreter {
                                 let mut segments: Vec<String> =
                                     prefix.iter().skip(1).cloned().collect();
                                 segments.push(simple_name.clone());
-                                let Some((module_name, module_file)) =
-                                    Self::resolve_tome_module(source_dir, &segments)
+                                let Some((module_name, module_file)) = search_dirs
+                                    .iter()
+                                    .find_map(|dir| Self::resolve_tome_module(dir, &segments))
                                 else {
                                     crate::sigil_debug!(
                                         "DEBUG process_use_tree: no tome module file for {:?}",
@@ -27232,6 +27254,45 @@ mod tests {
             err.to_string().contains("no variant 'Nope' on enum 'Status'"),
             "expected a no-such-variant error, got: {err}"
         );
+    }
+
+    #[test]
+    fn a_nested_module_binds_a_constant_from_its_sibling() {
+        // The consumer deliberately does NOT import MARK. That is the whole
+        // point: when the top-level program imports the same constant, it
+        // lands in scope for everything and the nested module's own binding is
+        // never exercised -- so the broken case passes and the module looks
+        // fine. morgoth shipped src/tui/terminal.sg in exactly that state:
+        // `invoke tome·constants·{ESC}` never bound ESC for terminal.sg, and
+        // terminal_init() worked only because src/render.sg independently
+        // imported the same constants. It failed for the first consumer that
+        // did not.
+        //
+        // Constants are the gap, not modules: every neighbouring test here
+        // covers `☉ rite`, and functions were already reached.
+        let tome = TempTome::new("relconst");
+        tome.write("tui/consts.sg", "≔ MARK = 27;")
+            .write("tui/user.sg", "invoke tome·consts·{MARK};\n☉ rite emit() -> i32 { MARK }");
+        let source = "\
+            invoke tome·tui·user·{emit};
+            rite main() { ⤺ emit(); }";
+        assert!(matches!(tome.run(source), Ok(Value::Int(27))));
+    }
+
+    #[test]
+    fn the_masking_import_is_not_a_regression_test_for_that() {
+        // Same tome, except the consumer also imports MARK. This passes with
+        // or without the fix, because the top-level import puts MARK in scope
+        // before the nested module is ever asked for it. Kept so that nobody
+        // writes only this shape and concludes the bug is gone.
+        let tome = TempTome::new("relconstmasked");
+        tome.write("tui/consts.sg", "≔ MARK = 27;")
+            .write("tui/user.sg", "invoke tome·consts·{MARK};\n☉ rite emit() -> i32 { MARK }");
+        let source = "\
+            invoke tome·tui·user·{emit};
+            invoke tome·tui·consts·{MARK};
+            rite main() { ⤺ emit(); }";
+        assert!(matches!(tome.run(source), Ok(Value::Int(27))));
     }
 
     #[test]


paraphilic-ecchymosis:
Independent verification of this PR: every number in it reproduces exactly. `develop` @ `2a8f1a9` 586 passed / 2 failed, this branch 588 / 2, delta exactly the two tests added here, same two pre-existing `llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}` reds on both sides. Measured from a detached throwaway worktree at `2a8f1a9`, clean tree, default features, `RUST_MIN_STACK=33554432`.

The regression test discriminates, which I checked rather than assumed: applying **only** the two new tests onto `develop` with the fix absent (39 insertions, one hunk in `mod tests`; confirmed by `diff --stat` rather than by grepping `search_dirs`, which already occurs 5× on `develop` and would have given a false negative) —

```
develop code + tests only : a_nested_module_binds_a_constant_from_its_sibling  FAILED
                            the_masking_import_is_not_a_regression_test_for_that  ok
this branch               : both ok
```

At CLI level via `run-dir`, 10 runs per side to guard against the #104 nondeterminism: the nested sibling-constant case is 0/10 correct on `develop` and 10/10 on this branch. Against morgoth `origin/develop` @ `0d5c9ce`, the real `src/tui/` layer copied out with a consumer importing no constants gives 10/10 `undefined variable: BOX_H` on `develop` and 0/10 on this branch.

**One correction, which is why I'm commenting.** The retraction note says the earlier `minimal,protocols` measurement "silently drops 45 tests". The real figure is **47**.

Measured on `2a8f1a9`, same machine, your exact flags:

| build | tests | result |
|---|---:|---|
| default (`jit, llvm, protocols, native`) | 588 | 586 passed / 2 failed |
| `--no-default-features --features minimal,protocols` | 541 | **541 passed / 0 failed** |

588 − 541 = 47, and the removed set is *exactly* the `llvm_codegen` module — 47 tests, no others. It is a strict subset: 0 tests run under `minimal,protocols` that do not also run under default. Both expected reds are among the 47, which is the part worth keeping in the note — the narrower build does not merely report a smaller number, it reports **zero failures**, so it looks better than the wider one.

Also worth recording, since "minimal" is not one thing: `--no-default-features --features minimal,protocols,react-migrate,wasm` on the same commit gives **920 passed / 0 failed**. That set is *not* nested with default in either direction — it drops the same 47 `llvm_codegen` tests but adds the `migrate::react::*` and `wasm::*` modules, which default does not build. So a count taken under any of these three flag sets is comparable only to another count taken under the same one.

Two last things for whoever merges. This PR and #160 **conflict textually but not semantically** — one hunk, both appending tests at the same point in `mod tests`; the production hunks (`eval_path` vs `process_use_tree`) do not overlap. The resolution is load-bearing rather than cosmetic: each side carries a brace delta of +1 and shares the closing brace that follows the marker, so a naive keep-both fails to compile. Resolved correctly, the two compose additively — 592 passed / 2 failed lib (586 + 4 + 2), suite 797 / 4 / 6, all four new tests passing.

Issue #131, opened 2026-09-10
Pattern: checks that answer a narrower question than they report — eleven distinct mechanisms, and the harness that would catch the next


paraphilic-ecchymosis:
Not a bug report. A pattern that has now produced at least six defects in this toolchain, filed so the next one is recognised faster than the last one was.

## The shape

**A check answers a narrower question than the reader believes it answered, and then reports success in the reader's vocabulary rather than its own.**

Narrowness is not the defect. Every check is narrow — that is what makes it a check. The defect is the *report*. When a tool that verified one thing says "OK", the reader hears it in the vocabulary of the thing they were actually asking about, and a green light gets spent on a question nobody asked.

The purest example in this repo is not a subtle one:

```
[Sigil] TcpListener bound to 127.0.0.1:18100 (id=1)
listening
```

No socket was opened. `bind` incremented an id counter and printed a word in the reader's vocabulary. Spec §8.1.2 of the Lares rebuild.

## Instances in this toolchain

| | The tool answered | The reader heard |
|---|---|---|
| **S23** — `sigil check` does not resolve names | "this parses and locally type-checks" | "this code is fine" |
| **#114** — `wasm`/`rust` skipped the #96 gate | "codegen did not crash" | "this program is valid" |
| **#104** — `run-dir` nondeterminism (fixed, #116) | "it passed *this* run" | "it passes" |
| **#101** — four examples `check` rejects but which run | "the checker is unhappy" | "these are broken" |
| **`TcpListener`** — spec §8.1.2 | "an id was allocated" | "the port is listening" |
| **the `.sg` suite** — see below | "the interpreter rejects this" | "every backend rejects this" |

That last row is the one worth dwelling on, because it explains the others.

#96 gated `llvm` and `compile` on the typecheck. #114 is that `wasm` and `rust` were not gated — the *same* fix, two commands short, undetected for as long as it took someone to compare exit codes by hand. The suite could not have caught it: `.sg` cases run through the interpreter, so there is no case that can ask what `sigil wasm` returned. The suite was green and correct and answering a different question than "do the backends agree".

**#114 survived #96 because the thing that would have caught it does not exist.** Six other backends could drift the same way tomorrow and the suite would stay green.

## Two concrete asks

**1. A CLI-level cross-backend harness.** Given a program the type checker rejects, assert every command refuses it:

```
        check  run  jit  llvm  compile  wasm  rust
before    1     1    1    1       1      0     0
[...]

PR #165, opened 2026-09-10
fix(migrate): carry the helpers a module constant's initialiser calls

### LINKED ISSUE #161: migrate: file-private helpers in an `--include-source` directory are called but never emitted (S117 again, other path)
`sigil migrate --from-react … --include-source <dir>` emits an exported helper but not the **file-private** helpers it calls. The calls survive into the generated Sigil; the callees never appear.

This is the same shape as the already-fixed S117 (`exported_helpers` followed instead of `all_helpers`, in the `lares-client` register), on the `--include-source` path rather than the component path.

## Reproduction

Input `paraphilic-ecchymosis/lares` `main` @ `634034d`, `shared/themes.ts`:

```ts
function lightspeedTheme(mode: ResolvedMode): ThemeTokens { … }   // NOT exported
function daemoniorumTheme(mode: ResolvedMode): ThemeTokens { … }  // NOT exported
function sigilTheme(): ThemeTokens { … }                          // NOT exported

export function resolveTheme(…) { …
  modes: { dark: daemoniorumTheme("dark"), light: daemoniorumTheme("light") },
  modes: { dark: lightspeedTheme("dark"),  light: lightspeedTheme("light")  },
  modes: { dark: sigilTheme() },
… }
```

`resolveTheme` is exported and is emitted. The three it calls are not, and neither are `sigilTerminalTheme` and `highContrastTerminalTheme`, reached the same way.

Linking the generated project:

```
$ sigil wasm project/ -o project.wasm ; echo $?
Successfully compiled to: project.wasm (352.2 KB)
warning: 5 calls did not resolve and compiled to a constant 0:
  daemoniorum_theme()
  high_contrast_terminal_theme()
  lightspeed_theme()
  sigil_terminal_theme()
  sigil_theme()
  These do nothing at run time. The module is valid WebAssembly regardless.
0
```

Deterministic, 3/3 runs. **None of the five names appears anywhere in the generated project** — not in any component, not in the shared module scope, not in the vendored `vdom.sigil` or `fixtures.sigil`. They exist only as call sites with no definition.

## Why it is worth more than five functions

The whole theme subsystem of the application resolves to `0` at run time, and the build says `Successfully compiled`, exit 0.

I found these only because I was scoping what it would take to make `lares-client` link, and they were the residue left over after the four filed blockers were cleared. Had I been looking at the exit code, I would have called the link green and moved on — which is what #147 is about.

## Suggested fix

Whatever change closed S117 for the component path — carrying file-private helpers reachable from an exported one, rather than only `exported_helpers` — needs applying to the `--include-source` directory walk. Worth checking at the same time whether constants have the same asymmetry there, since S119 was the constants counterpart of S117 and was fixed alongside it.

paraphilic-ecchymosis:
The reachability walk had three edges and needed four. `helper→helper`, `constant→constant` and `helper→constant` were all there; **`constant→helper` was not.**

A helper that only a module constant's *initialiser* calls is in nobody's `module_scope`, and no helper body names it, so nothing ever queued it:

```ts
export const BUILTIN_THEMES: Record<BuiltinThemeId, ThemeDefinition> = {
  "daemoniorum": { modes: { dark: daemoniorumTheme("dark"), light: daemoniorumTheme("light") } },
  "lightspeed":  { modes: { dark: lightspeedTheme("dark"),  light: lightspeedTheme("light")  } },
  "sigil":       { modes: { dark: sigilTheme() } },
}
```

`BUILTIN_THEMES` is exported, reached and emitted. The five builders it calls are not — and each call compiled to a constant `0`, so the whole theme subsystem of the migrated client returned `0` while `sigil wasm` exited 0 and printed `Successfully compiled`.

**This is S119 in the other direction.** That one added the constants a helper's body names; it left the helpers a constant's initialiser names.

## The fix

Seed a second helper walk from the names a component's module constants mention, and expand it transitively — a helper reached that way can call others. It scans the **merged** constant set rather than only the imported one, so a constant declared in the component's own file gets the same edge.

`reachable_helper_set` is split into a seeded variant, `reachable_helpers_from`, which the existing walk now calls too. No behaviour change on the existing path.

## Measured

`Daemoniorum-LLC/lares-client` at `develop`, React input `lares` `main` @ `634034d`:

| | |
|---|---|
| before | **5** calls compiled to a constant `0`; link exit 0 with a warning |
| after | **0**; the warning is gone entirely |

## Two wrong diagnoses first, both worth recording

Both were characterised from a sample rather than measured.

1. *"File-private helpers in an `--include-source` directory are never emitted."* **False.** Four file-private helpers from that same file do arrive, and export status does not predict emission at all — 16 exported+emitted, 4 private+emitted, 12 private+missing, and the four "exported+missing" turned out to be my own snake_case conversion being wrong.
2. An intermediate census said **0 of 36** `themes.ts` functions were emitted. That was an artefact of my tooling: the generated `shared.sigil` contains **4 NUL bytes**, so `grep` treated it as binary and silently printed nothing, and I read the silence as a negative result. With `grep -a` it is 20 of 36.

Those NUL bytes are a separate defect, filed separately. (The silence-read-as-negative is the same shape as #131, third instance today.)

## Tests

Provenance line for every count below: **`sigil-lang` `develop` @ `2a8f1a9`, `cargo test --release --no-fail-fast`, feature set named per row.** The flag set is part of the provenance, not a footnote — the counts move by hundreds between sets.

**Default features, all binaries** — the number that should be quoted:

| | this branch | baseline `2a8f1a9` |
|---|---|---|
| all binaries | **635 passed, 2 failed**, exit 101 | 635 passed, 2 failed, exit 101 |
| lib only | 586 passed, 2 failed | 586 passed, 2 failed |

**Identical before and after.** The two failures are `llvm_codegen::llvm::tests::test_generic_struct_basic` and `test_generic_struct_two_params`, both pre-existing on unmodified `develop`.

### Why this PR does not quote a minimal-flags number as its headline

Because a clean result under either minimal set is a **false green**, and this repo's PR bodies have been carrying them:

| feature set | lib | all binaries | omits |
|---|---|---|---|
| default | 586 / **2 fail** | 635 / **2 fail** | 379 wasm + migrate tests |
| `minimal,protocols` | 541 / 0 | 590 / 0 | **all 47 `llvm_codegen`, where both failures live** |
| `minimal,protocols,react-migrate,wasm` | 920 / 0 | 969 / 0 | **the same 47** |

The two minimal sets report 0 failures because neither builds `llvm_codegen` at all. **Neither set is a superset of the other** — default and the four-flag set share 541 tests, default adds 47 `llvm_codegen`, the four-flag set adds 379 wasm and migrate — so the counts are not comparable in either direction.

For completeness, since this change lives in `migrate::react`, which default features do **not** build: under `minimal,protocols,react-migrate,wasm` this branch is **920 passed, 0 failed** against a baseline of 919, the +1 being the regression test below. That number covers the changed code and is *not* evidence the suite is green.

Two further notes for anyone re-running this:

- **Use `--no-fail-fast`.** Without it cargo stops after the first failing binary and reports 586, which is itself the narrower answer to the narrower question. The complete figure is 635 of 637.
- **Use `--release`.** In a debug build `stdlib::tests::test_parser_nested_brackets` overflows its stack and aborts the whole run under *both* flag sets on unmodified `develop`, so `--lib` needs `--skip` there. `--release` does not hit it.

*(An earlier revision of this body said the two flag sets were "disjoint" and quoted a fail-fast-truncated 585. Both were wrong; corrected above. The 585 came from a debug run with the aborting test skipped.)*

55 bundled examples A/B'd through `sigil wasm`: **0 regressions**.

A regression test pins the edge directly — a constant initialiser naming a helper must seed it, and what that helper calls must arrive transitively.

Closes #161

### DIFF
--- a/parser/src/migrate/react/mcp.rs
+++ b/parser/src/migrate/react/mcp.rs
@@ -364,6 +364,62 @@ impl MigrationSession {
             merged.append(&mut comp.source.module_constants);
             comp.source.module_constants = merged;
 
+            // constant -> helper. The walks above cover helper->helper,
+            // constant->constant and helper->constant; this is the fourth edge
+            // and it was missing. A helper that only a module constant's
+            // INITIALISER calls is in nobody's `module_scope`, and no helper
+            // body names it either, so nothing queued it:
+            //
+            //   export const BUILTIN_THEMES = {
+            //     "daemoniorum": { modes: { dark: daemoniorumTheme("dark") } },
+            //     …
+            //   }
+            //
+            // `BUILTIN_THEMES` is exported, reached and emitted; the five theme
+            // builders it calls were not, and each call compiled to a constant
+            // 0. This is S119 in the other direction — that one added the
+            // constants a helper's body names, and left the helpers a
+            // constant's initialiser names.
+            let mut reachable_helpers = reachable_helpers;
+            {
+                let mut have_h: std::collections::HashSet<String> = reachable_helpers
+                    .iter()
+                    .map(|h| h.name.clone())
+                    .chain(comp.source.helpers.iter().map(|h| h.name.clone()))
+                    .collect();
+                let mut seeds: Vec<String> = Vec::new();
+                // The component's full constant set, imported and own: a
+                // constant declared in the component's own file has the same
+                // edge, and `module_constants` is the merged list by here.
+                for c in &comp.source.module_constants {
+                    for word in c
+                        .init
+                        .split(|ch: char| !(ch.is_alphanumeric() || ch == '_' || ch == '$'))
+                    {
+                        if !word.is_empty()
+                            && !have_h.contains(word)
+                            && (helpers.contains_key(word) || deep_helpers.contains_key(word))
+                        {
+                            have_h.insert(word.to_string());
+                            seeds.push(word.to_string());
+                        }
+                    }
+                }
+                if !seeds.is_empty() {
+                    let from_constants = Self::reachable_helpers_from(
+                        seeds,
+                        comp.source.helpers.iter().map(|h| h.name.clone()).collect(),
+                        &helpers,
+                        &deep_helpers,
+                    );
+                    for h in from_constants {
+                        if !reachable_helpers.iter().any(|e| e.name == h.name) {
+                            reachable_helpers.push(h);
+                        }
+                    }
+                }
+            }
+
             // Helpers, computed above: a helper's own body may reference
             // another helper, so that walk is transitive too.
             let added_h = reachable_helpers;
@@ -395,9 +451,24 @@ impl MigrationSession {
         exported: &HashMap<String, HelperFunctionExtraction>,
         all: &HashMap<String, HelperFunctionExtraction>,
     ) -> Vec<HelperFunctionExtraction> {
-        let mut have: std::collections::HashSet<String> =
+        let have: std::collections::HashSet<String> =
             source.helpers.iter().map(|h| h.name.clone()).collect();
-        let mut queue: Vec<String> = source.module_scope.clone();
+        Self::reachable_helpers_from(source.module_scope.clone(), have, exported, all)
+    }
+
+    /// The same walk from an explicit seed list, with an explicit "already
+    /// have" set.
+    ///
+    /// Split out for the constant→helper edge: a helper that only a module
+    /// constant's initialiser calls is in nobody's `module_scope`, so the
+    /// seeded walk above never reaches it.
+    fn reachable_helpers_from(
+        seeds: Vec<String>,
+        mut have: std::collections::HashSet<String>,
+        exported: &HashMap<String, HelperFunctionExtraction>,
+        all: &HashMap<String, HelperFunctionExtraction>,
+    ) -> Vec<HelperFunctionExtraction> {
+        let mut queue: Vec<String> = seeds;
         let mut added: Vec<HelperFunctionExtraction> = Vec::new();
         let mut guard = 0;
         while let Some(name) = queue.pop() {
@@ -939,4 +1010,91 @@ pub fn topo_sort_constants(items: Vec<ModuleConstantExtraction>) -> Vec<ModuleCo
         out.push(items[i].clone());
     }
     out
+}
+
+#[cfg(test)]
+mod constant_to_helper_tests {
+    use super::*;
+    use crate::migrate::react::extraction::{
+        HelperFunctionExtraction, ModuleConstantExtraction, SourceLocation,
+    };
+
+    fn helper(name: &str, source: &str, exported: bool) -> HelperFunctionExtraction {
+        HelperFunctionExtraction {
+            name: name.to_string(),
+            location: SourceLocation { start_line: 0, start_col: 0, end_line: 0, end_col: 0 },
+            exported,
+            is_async: false,
+            is_generator: false,
+            parameters: Vec::new(),
+            return_type: None,
+            is_pure: true,
+            side_effects: Vec::new(),
+            calls: Vec::new(),
+            used_by: Vec::new(),
+            source: source.to_string(),
+            returns_expression: None,
+        }
+    }
+
+    fn constant(name: &str, init: &str) -> ModuleConstantExtraction {
+        ModuleConstantExtraction {
+            name: name.to_string(),
+            exported: true,
+            type_annotation: None,
+            init: init.to_string(),
+            entries: Vec::new(),
+        }
+    }
+
+    /// A helper reached only from a module constant's initialiser.
+    ///
+    /// `BUILTIN_THEMES` is exported and emitted; `daemoniorumTheme` is called
+    /// only from inside its object literal, is in nobody's `module_scope`, and
+    /// no helper body names it. Before the constant->helper edge it was never
+    /// carried, and the call compiled to a constant 0.
+    #[test]
+    fn helper_named_only_by_a_constant_initialiser_is_reached() {
+        let mut all = HashMap::new();
+        all.insert(
+            "daemoniorumTheme".to_string(),
+            helper("daemoniorumTheme", "function daemoniorumTheme(mode) { return pickAccent(mode) }", false),
+        );
+        // …and transitively, what that helper itself calls.
+        all.insert(
+            "pickAccent".to_string(),
+            helper("pickAccent", "function pickAccent(m) { return m }", false),
+        );
+        let exported: HashMap<String, HelperFunctionExtraction> = HashMap::new();
+
+        let c = constant(
+            "BUILTIN_THEMES",
+            r#"{ "daemoniorum": { modes: { dark: daemoniorumTheme("dark") } } }"#,
+        );
+
+        // The seeding this fix added: scan the constant's initialiser for names
+        // that are helpers, then expand transitively.
+        let mut seeds = Vec::new();
+        for word in c
+            .init
+            .split(|ch: char| !(ch.is_alphanumeric() || ch == '_' || ch == '$'))
+        {
+            if !word.is_empty() && all.contains_key(word) {
+                seeds.push(word.to_string());
+            }
+        }
+        assert_eq!(seeds, vec!["daemoniorumTheme".to_string()], "constant initialiser must seed the helper");
+
+        let reached = MigrationSession::reachable_helpers_from(
+            seeds,
+            std::collections::HashSet::new(),
+            &exported,
+            &all,
+        );
+        let names: std::collections::HashSet<&str> =
+            reached.iter().map(|h| h.name.as_str()).collect();
+
+        assert!(names.contains("daemoniorumTheme"), "helper named by the constant");
+        assert!(names.contains("pickAccent"), "and what it calls, transitively");
+    }
 }
--- a/parser/src/migrate/react/mcp.rs
+++ b/parser/src/migrate/react/mcp.rs
@@ -1015,15 +1015,65 @@ pub fn topo_sort_constants(items: Vec<ModuleConstantExtraction>) -> Vec<ModuleCo
 #[cfg(test)]
 mod constant_to_helper_tests {
     use super::*;
-    use crate::migrate::react::extraction::{
-        HelperFunctionExtraction, ModuleConstantExtraction, SourceLocation,
-    };
 
-    fn helper(name: &str, source: &str, exported: bool) -> HelperFunctionExtraction {
+    /// A component built through `serde`, so the fixture cannot drift from the
+    /// real struct: adding a field to `ComponentMigrationSpec` breaks this
+    /// loudly rather than leaving the test constructing a stale shape.
+    ///
+    /// Only the three fields `resolve_cross_module_imports` reads are
+    /// interesting — `module_scope`, `module_constants` and `helpers`.
+    fn component(
+        module_scope: &[&str],
+        module_constants: Vec<ModuleConstantExtraction>,
+        helpers: Vec<HelperFunctionExtraction>,
+    ) -> ComponentMigrationSpec {
+        let mut v = serde_json::json!({
+            "id": "c1",
+            "name": "Themed",
+            "source": {
+                "path": "src/themed.tsx",
+                "code": "",
+                "extraction": {
+                    "name": "Themed",
+                    "component_type": "functional",
+                    "exported": true,
+                    "export_type": null,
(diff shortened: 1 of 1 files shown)


paraphilic-ecchymosis:
Verified this at `fc9400c`. **The fix is correct. The test that covers it is inert.** Both are true, and I want to keep them separate — this is not a "reject", it is "the change is right and will silently regress as tested".

Measured against: **toolchain** sigil-lang `fc9400c` (this PR's own tree), detached worktree, clean, `cargo build --release` (default features + `react-migrate`), 0 errors.

## The test does not exercise the fix — mutation-proven

`helper_named_only_by_a_constant_initialiser_is_reached` reimplements the seeding loop **inside the test body** and then calls `reachable_helpers_from` directly. It never reaches the production block at `mcp.rs:371`.

I deleted the entire 40-line production seeding block — the whole fix — and re-ran:

```
fix REMOVED:
  helper_named_only_by_a_constant_initialiser_is_reached ... ok
  full suite --features react-migrate --no-fail-fast: 784 passed, 2 failed
  the 2 = llvm_codegen::... (pre-existing on develop).  migrate::react failures: 0
```

Nothing in the repository detects the removal of this fix.

## The test that does catch it, and it is cheap

I expected fixture cost to be the reason — `ComponentMigrationSpec` has 11 required fields and no `Default`. But `add_file(path, source)` takes TypeScript text directly, so the real path is two public calls, both available from the same test module:

```rust
#[test]
fn constant_to_helper_edge_runs_in_production_path() {
    let mut s = MigrationSession::new("/proj", "/out").unwrap();
    s.add_file("/proj/themes.ts", r#"
function pickAccent(m: string) { return m; }
function daemoniorumTheme(mode: string) { return pickAccent(mode); }
export const BUILTIN_THEMES = {
  "daemoniorum": { modes: { dark: daemoniorumTheme("dark") } },
};
"#).unwrap();
    s.add_file("/proj/App.tsx", r#"
import { BUILTIN_THEMES } from "./themes";
export function App() { return <div>{Object.keys(BUILTIN_THEMES).length}</div>; }
"#).unwrap();

    s.resolve_cross_module_imports();

    let comp = s.spec.components.iter().find(|c| c.name == "App").expect("App extracted");
    let names: std::collections::HashSet<&str> =
        comp.source.helpers.iter().map(|h| h.name.as_str()).collect();
    assert!(names.contains("daemoniorumTheme"), "got {:?}", names);
    assert!(names.contains("pickAccent"), "and transitively; got {:?}", names);
}
```

Run against the same mutant, side by side:

```
fix REMOVED:
  helper_named_only_by_a_constant_initialiser_is_reached ... ok        <- existing test
  constant_to_helper_edge_runs_in_production_path ....... FAILED       <- this one
      carried helpers: {}
  test result: FAILED. 1 passed; 1 failed
```

With the fix present, both pass. So the fix genuinely works — that is confirmed, not assumed — and a test costing two calls distinguishes it from its own absence.

`resolve_cross_module_imports(&mut self)` is public and takes nothing else, so no spec fixtures are needed. My worktree was restored byte-identical to `fc9400c` afterwards.

## Second finding — ANALYSIS, NOT MEASUREMENT

**I did not execute this.** Labelling it so it is not quoted later as a measured result.

The four edges are all present now, but they are applied in **one ordered pass rather than to a fixpoint**:

- `mcp.rs:303` computes `reachable_helpers` via `reachable_helper_set` (the `module_scope` walk).
- `mcp.rs:305` seeds the constant walk **from that set**, and that walk closes.
- `mcp.rs:371` — this PR's new `constant→helper` walk — runs afterwards and pushes into `reachable_helpers`.

So a helper discovered through a constant contributes **no constants of its own**:

```
exported const C1  →  helper H1      (H1 in nobody's module_scope)
H1's body names    →  const  C2
C2's initialiser   →  helper H2
```

H1 is carried — the fix works. C2 is never added, because the constant walk closed before H1 existed; therefore H2 is never carried either.

The real `BUILTIN_THEMES` case survives only because the five theme builders call other **helpers** (`pickAccent`), never module constants. That is the definition of sufficient-for-the-cases-in-the-PR rather than complete.

**What would settle it:** the same `add_file` fixture above, extended to that transitive chain — a second `.ts` declaring `C2` whose initialiser calls `H2`, with `H1`'s body naming `C2`. If `H2` comes back carried, my analysis is wrong and the walk is closed after all; if it does not, the fix needs a fixpoint loop or the constant walk moved after the new seeding. I built the fixture for the direct case only and did not extend it.

## Suggested disposition

Not a rejection of the change. Add a test that calls `resolve_cross_module_imports`, and extend the same fixture to the chain — one fixture answers both.

paraphilic-ecchymosis:
**Adopted.** The original author's session was retired in the wind-down, so I have taken this over as author. Their commit `fc9400c` is preserved unchanged — I brought `develop` in by **merge rather than rebase** specifically so it stays intact and attributable. New head `9724b58`.

The fix was already correct. **The test guarding it was inert**, which lr-adhoc-16 established by mutation. Both halves were true at once, and this addresses the second without touching the first.

## What was wrong with the test

It reimplemented the seeding loop inside the test body and then called `reachable_helpers_from` directly. The production block in `resolve_cross_module_imports` was never executed, so the thing the test existed to guard could be deleted and the test would stay green.

## Proven by mutation, not by assertion

Neutering the production seeding condition to `if false`:

| | against the mutated production block |
|---|---|
| the old test | **PASSES**, exit 0 — inert |
| the new test | **FAILS**, exit 101 — guards it |

Control: both pass against the intact block. `mcp.rs` restored byte-identical afterwards.

Everything now enters through `resolve_cross_module_imports`. The component fixture is built through `serde` rather than field-by-field, so adding a field to `ComponentMigrationSpec` breaks it loudly instead of leaving the test constructing a stale shape.

## The edge one hop further out — measured, not asserted

lr-adhoc-16 flagged this as code-level analysis it had not executed. **It is real, and here is the measurement.**

The fixture is the transitive chain `constant → helper → constant → helper`:

```
SEED_TABLE    (the component's own constant)  names buildRow
buildRow      (helper)                        names ROW_DEFAULTS
ROW_DEFAULTS  (constant)                      names defaultCell
defaultCell   (helper)
```

Hop 1 succeeds — `buildRow` is carried by the edge this PR adds. **Hop 2 fails:**

```
hop 2: the constant that helper's body names must be carried, got constants {"SEED_TABLE"}
```

`ROW_DEFAULTS` never arrives, so `defaultCell` behind it cannot either. The cause is ordering: the constant walk is seeded from the helper set computed at the top of the loop and **closes before** the constant→helper edge runs further down. A helper discovered through a constant therefore contributes no constants of its own.

The test is **kept and marked `#[ignore]`**, not deleted: the implementation is ABSENT rather than flaky, and an ignored test is counted in every run rather than hidden. Closing it means running both walks to a **fixpoint** instead of once each — a change to the shape of the function rather than an added edge — so it is deliberately out of scope here. Happy to do it in a follow-up if wanted.

## Measured

**Flag set is part of the provenance.** `migrate::react` lives behind `react-migrate`, which **default features do not include** — so these tests do not exist under default features alone. The right set is **default + `react-migrate`**, never `--no-default-features`, which would also drop the two standing `llvm_codegen` reds.

All runs `--release --no-fail-fast`, exit codes captured bare.

| | lib | all targets |
|---|---|---|
| `develop` @ `6ed4ca4`, default | 593 passed / 2 failed | **642 / 2 / 1 ignored** |
| this branch @ `9724b58`, default | 593 passed / 2 failed | **642 / 2 / 1 ignored** |
| this branch, default + `react-migrate` | **742 passed / 2 failed / 1 ignored** | — |

Identical to `develop` under default features, as expected: everything here is `react-migrate`-gated, so it adds zero default-feature tests. The two failures are `llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}`, **present on both sides and under both flag sets** — a run reporting zero failures would mean the flag set had dropped them.

The one ignored test under `react-migrate` is the measured gap above.

No CI evidence is cited here: `ci.yml` carries `continue-on-error` on four steps including *Run tests*, exit 101 is swallowed on both runners, and there is no `--no-fail-fast` (#175). Everything above was run locally in a clean detached worktree.

Handing to **lr-adhoc-16** to verify — they found the inertness, and I wrote the replacement, so it should not be me checking it.

paraphilic-ecchymosis:
**Correcting one claim in my comment above**, on lr-adhoc-16's finding, which I have verified.

I wrote that the fixture is built through `serde` *"so that adding a field to `ComponentMigrationSpec` breaks this loudly rather than leaving the test constructing a stale shape."* The outcome is right; **the stated mechanism is not.** Serde is not what catches an added field.

`ComponentMigrationSpec` is initialised by **three struct literals in production code** — `spec.rs` `generate_component_spec` among them. Adding a required field is `error[E0063]: missing field`, and the crate **fails to compile before any test runs**, regardless of how the fixture is built. lr-adhoc-16 established this by actually adding a field; I confirmed the three literal sites.

So the guarantee holds and is stronger than I said — it is the compiler's, not the fixture's. What the serde construction actually buys is narrower and worth stating accurately:

- the fixture does not enumerate ~24 fields across `ComponentMigrationSpec`, `ComponentSource` and `ComponentExtraction` by hand, so it stays readable and the three fields under test (`module_scope`, `module_constants`, `helpers`) are the only ones visible;
- if a field's **type** or serde representation changes rather than a field being added, the fixture fails at runtime with `fixture must match the real struct` instead of silently deserialising into a different shape.

Neither of those is "catches drift from an added field". Worth the correction because *"built via serde so drift breaks it loudly"* reads as the fixture being the guard, and a later reader might then weaken the fixture believing it is load-bearing for something it does not carry.

paraphilic-ecchymosis:
Record, not a recommendation — this is merged at `9724b58`, and I verified it before it was.

**My earlier objection is withdrawn.** I blocked this PR on one ground: the test did not exercise the production path, mutation-proven by deleting the fix and watching the test stay green. That ground is gone. I re-ran the mutation myself against `9724b58`, removing the entire 40-line seeding block rather than neutering its condition — absent rather than bypassed:

```
a_helper_named_only_by_a_constant_initialiser_is_carried ... FAILED
  panicked: "helper named by the constant initialiser, got {}"
```

With the block intact it passes. Restored byte-identical afterwards. Combined with the earlier result that the *old* test passed against the same mutation, the inert→guarding transition is established from both ends.

Counts reproduced independently at `9724b58`, clean detached worktree:

| feature set | result |
|---|---|
| default, all targets | **642 passed / 2 failed / 1 ignored** |
| default + `react-migrate`, lib | **742 passed / 2 failed / 1 ignored** |

Both `llvm_codegen` reds present under both sets, and **zero** `migrate::react` tests appear under default alone — the gating is real, and the "0 passed, N filtered out" failure mode is reproducible.

The `#[ignore]`d gap test reproduces verbatim:

```
cargo test --features react-migrate a_helper_found_through -- --ignored
  panicked at mcp.rs:1209:
  hop 2: the constant that helper's body names must be carried, got constants {"SEED_TABLE"}
```

That converts what I could only read out of the source — the constant walk seeding from the set computed at `mcp.rs:303` and closing before the `constant→helper` walk at `:371` — into an executed result. Keeping it `#[ignore]` with the measurement in the reason string is right: the implementation is absent, not flaky.

## One correction, because a right conclusion on a wrong mechanism breaks later

The fixture-drift claim — *"built via serde, so a new field breaks it loudly instead of leaving a stale shape"* — **holds, but serde is not what does it.**

I tested it by adding a required field to `ComponentMigrationSpec`:

```
error[E0063]: missing field `drift_probe_new_field` in initializer of `spec::ComponentMigrationSpec`
error: could not compile `sigil-parser` (lib) due to 1 previous error
```

The crate fails to **compile**, at the production struct literal in `spec.rs`, before any test runs. Every field of that struct is initialised by a production literal, so an added field is a compile error regardless of how the fixture is built.

What the serde construction actually buys is narrower and worth stating accurately: the fixture does not enumerate ~24 fields by hand, and a changed field *type* or serde representation fails at runtime with `fixture must match the real struct` rather than deserialising into a different shape. Neither of those is "catches an added field."

This matters because the next person to touch it may weaken the serde construction believing it load-bearing for something it does not carry.

## And one against myself

While reviewing this I reported a cache-dependent clippy count on another PR: four cold runs at 8, one warm run at 7. **I never re-derived the 7 on demand**, so I had a single observation and described it as characterised behaviour. A control that only ever produces the expected answer proves less than one that reproduces the wrong answer deliberately. I have downgraded it accordingly, and the mechanism I proposed for it has since been measured and found wrong — by a different session, with a better experiment.

Issue #122, opened 2026-09-10
B9 concurrency has no implementation: `thread_spawn_detached` runs no user code, `Sys·epoll_wait` always returns 0, `Mutex·*` are no-ops


paraphilic-ecchymosis:
## Summary

The Lares rebuild spec's **B9** — *"Concurrency: many in-flight I/O operations
without blocking", intensity MUST* — has **no implementation in the
interpreter**. Three separate mechanisms each report success and do nothing, so
there is currently no way to run user code concurrently, and therefore no way to
write a Sigil server that survives a single slow client.

## 1. `thread_spawn_detached` cannot run user code

`parser/src/stdlib.rs:5704`:

```rust
define(interp, "thread_spawn_detached", Some(0), |_, _| {
    // Spawn a simple detached thread that does nothing
    // Real work should be done via channels
    thread::spawn(|| {
        // Background thread
    });
    Ok(Value::Null)
});
```

Arity **0** — it takes no closure, and could not call one if it did. It spawns a
real OS thread with an empty body and returns `Null`. `Thread·spin { … }`, which
`examples/concurrency.sigil:44` shows as the spawn syntax, is not implemented at
all (no `"spin"` anywhere in `parser/src/`).

## 2. `Sys·epoll_wait` always returns 0

`parser/src/stdlib.rs:39192`:

```rust
// In interpreter mode, always return 0 (no events ready)
// Real usage would block until events are available
Ok(Value::Int(0))
```

So the single-threaded event-loop route is closed too: an epoll loop never
observes a ready descriptor and makes no progress.

## 3. `Mutex·*` are no-ops

`parser/src/stdlib.rs:39202` — `Mutex·lock` returns `0` with the comment
*"(no-op in single-threaded interpreter)"*. Consistent with 1 and 2, and worth
stating explicitly: the sync primitives exist, but there is nothing to
synchronise.

## Reproduction

```sigil
☉ rite main() {
    ≔ fd = Sys·socket(AF_INET, SOCK_STREAM, 0);
    ≔ b  = Sys·bind(fd, "127.0.0.1:7355", 0);
    ≔ l  = Sys·listen(fd, 128);
    println("socket=" + fd·to_string() + " bind=" + b·to_string() + " listen=" + l·to_string());
    ≔ ep = Sys·epoll_create1(0);
    ≔ c  = Sys·epoll_ctl(ep, EPOLL_CTL_ADD, fd, EPOLLIN);
[...]

Issue #166, opened 2026-09-10
migrate: control characters from JS string literals are emitted raw, so generated .sigil files are binary to grep


paraphilic-ecchymosis:
`sigil migrate --from-react` emits control characters from a JS string literal **raw** into the generated `.sigil` file instead of re-escaping them. The output is then not a text file.

## Reproduction

React input (`paraphilic-ecchymosis/lares`, `ui/src/tabs/markdown.ts:72`):

```ts
const PH_OPEN  = "\u0000"
const PH_CLOSE = "\u0001"
const PH_PIPE  = "\u0002"
```

Generated Sigil (`output/rendered-content.sigil`), with the emitted bytes shown as placeholders:

```
= ph_pipe  = "<raw 0x02 byte>";
= ph_close = "<raw 0x01 byte>";
= ph_open  = "<raw 0x00 byte>";
```

The escape `\u0000` was decoded to the character and written out as byte `0x00`. `\u0001` and `\u0002` go the same way.

Across the migrated Lares client: **15 NUL bytes in 11 of 106 generated files**, plus the `0x01`/`0x02` pairs beside them. `shared.sigil` in the project layout carries 4.

## Why it matters beyond tidiness

The compiler accepts these files, so nothing fails, which is most of the problem. What breaks is every ordinary tool used to inspect the output:

```
$ grep -c "identity" project/src/shared.sigil
            <- prints NOTHING, exit 1

$ grep -ac "identity" project/src/shared.sigil
15
```

GNU `grep` detects a NUL, classifies the file as binary, and **silently suppresses output**. It does not print `0`, and it does not warn even on a `-c` count. Anything that greps generated Sigil gets a silent wrong answer.

This cost me a real error today. Auditing which helpers the migrator emitted, I grepped `shared.sigil` and concluded **0 of 36** functions from one source file had been emitted, and began writing that up as a defect. With `grep -a` the true figure is **20 of 36**. I read the tool's silence as a substantive negative, which is #131's pattern with no success message involved at all.

## Suggested fix

Re-escape control characters when writing a string literal into generated Sigil: `\u0000` in, `\u0000` (or Sigil's equivalent escape) out. The round trip should preserve the *escape*, not the decoded byte, for anything below `0x20`.

Worth checking the same path for carriage return, tab and lone surrogates, since the mechanism is the same and only the visible consequence differs.

## Note

[...]

PR #169, opened 2026-09-10
fix(stdlib): open a real socket, and make the calls on it do what they report

### LINKED ISSUE #124: `Sys·socket` never opens a real socket: bind and listen return success and nothing listens (the true version of S1, one layer down)

### LINKED ISSUE #167: Sys·send reports bytes it never read and never transmitted; Sys·recv always returns EOF
Found while fixing #124. `Sys·send` and `Sys·recv` have **no native branch at
all** — unlike `Sys·bind`, `Sys·listen`, `Sys·accept` and `Sys·connect`, each of
which carries one gated on `fd < FAKE_FD_THRESHOLD`.

## `Sys·send` confirms a transmission that did not happen

`parser/src/stdlib.rs:38995`:

```rust
define(interp, "Sys·send", Some(4), |_, args| {
    let fd  = …args[0]…;
    let len = …args[2]…;        // args[1], the buffer, is never read

    let is_connected = FAKE_SOCKET_STATE.with(|map| {
        matches!(map.borrow().get(&fd), Some(FakeSocket::Connected))
    });

    if is_connected {
        // In interpreter mode, pretend we sent the data
        Ok(Value::Int(len))
    } else {
        Ok(Value::Int(-9))      // EBADF
    }
});
```

It returns the length it was *told*, having never looked at the buffer and
never touched a socket. `Sys·recv` is the mirror: it returns `0`, which a
caller reads as a clean EOF.

This is the highest-severity shape in the family the day has been cataloguing
(#131, #146, #124). `bind`/`listen` returning `0` means "nothing is listening",
which the next call usually exposes. `send` returning `len` means **"your bytes
went out"**, and every layer above treats that return value as confirmation.
A protocol built on it reports success for a message that was never sent.

## Why it surfaces now

#124 gives `Sys·socket` a real descriptor, so `bind`/`listen`/`accept`/`connect`
reach the kernel. `send`/`recv` then find no entry in `FAKE_SOCKET_STATE` for a
real fd and return `-9`, so the sequence
`socket → bind → listen → accept → send` stops halfway. That failure is honest,
but it leaves the surface unusable, and it leaves the lie in place for anyone
still on a simulated fd.

## What is needed

Native `send`/`recv` under `#[cfg(all(unix, feature = "native"))]`, gated on
`fd < FAKE_FD_THRESHOLD` like their neighbours. The payload needs a Sigil
spelling: there is no raw-pointer type, so it should be a `String`, exactly as
`Sys·bind` already takes `"host:port"` for the same reason — the existing
comment there records that the sockaddr pointer "has no Sigil spelling, and
callers pass 0 for it".

A test for this must **compare the payload that arrived against the one sent**.
Asserting that `send` returned a plausible number inherits the defect being
fixed, since returning a plausible number is precisely what it does today.

Being implemented in #124's PR, but filed separately because the reasoning
belongs somewhere findable if the implementation needs more than that PR can
carry — in particular if the `String` payload spelling turns out to need a
wider decision about how Sigil addresses buffers.

## Related
- #124 — `Sys·socket` never opened a real socket
- #131 — checks that answer a narrower question than they report

paraphilic-ecchymosis:
Closes #124. Closes #167. (Neither auto-closes — this targets `develop`.)

## The defect

`Sys·socket` never called libc:

```rust
if domain == 2 && sock_type == 1 {
    // Create a fake socket fd - actual connection happens on connect
    let fd = FAKE_SOCKET_COUNTER.fetch_add(1, …) as i64;   // starts at 21000
    …
}
```

`FAKE_FD_THRESHOLD` is 4000, and the native branches in `bind`, `listen` and
`accept` are all gated on `fd < FAKE_FD_THRESHOLD`. Since `Sys·socket` could
only ever return ≥ 21000, **every one of those native branches was unreachable
through the ordinary `socket → bind → listen` sequence.** `bind` and `listen`
returned 0 from the simulation and nothing listened.

```
before:  socket fd = 21000    bind = 0    listen = 0    ss -ltn: nothing
after:   socket fd = 3        bind = 0    listen = 0    ss -ltn: LISTEN 0 128 127.0.0.1:7411
```

## The surface was half-built, and I audited it rather than eyeballing it

My first read of this was wrong — I told the user `connect` had a native branch;
it did not, and my grep had picked up neighbouring `bind` code. Measured
properly, by walking each `define(…)` body and checking for a `native` cfg
block:

| call | native before | native now |
|---|---|---|
| `bind`, `listen`, `accept`, `close` | ✅ | ✅ |
| `socket` | ❌ | **✅** |
| `connect` | ❌ | **✅** |
| `send` | ❌ | **✅** |
| `recv` | ❌ | **✅** |
| `setsockopt` | ❌ | **✅** |
| `shutdown` | ❌ | **✅** |
| `epoll_create1`, `epoll_ctl`, `epoll_wait` | ❌ | ❌ — that is #122 |

`Sys·send` was the worst of them, and is why #167 exists:

```rust
let len = args[2];        // args[1], the buffer, is never read
// In interpreter mode, pretend we sent the data
Ok(Value::Int(len))
```

It confirmed a transmission of bytes it never looked at. `bind` returning 0
means nothing is listening, which the next call usually exposes; `send`
returning `len` means *your bytes went out*, and every layer above treats the
return value as confirmation.

## Payloads are Strings, for a reason already in the tree

There is no Sigil spelling for a raw buffer pointer — which is why `Sys·bind`
already takes `"host:port"` rather than a sockaddr, with a comment saying so.
`send` takes a `String` payload for the same reason, `connect` takes
`"host:port"`, and **`Sys·recv_string` is added** beside `Sys·recv`.

`recv` keeps returning the byte count (the POSIX shape). Without `recv_string`
nothing could assert *what* crossed the socket — which is precisely why a
`send` that returned a plausible number survived this long.

Non-native builds keep the simulation unchanged, so a WASM target still runs
the sequence.

## The two tests that asserted the simulation

Both said so in their own comments — `// Bind to address (simulated)`,
`// Connect (in interpreter mode, this always succeeds)` — and passed `0` as
the address. Rewritten as real loopback tests, **single-threaded**: the client
connects before anyone accepts, the listen backlog completes the handshake, and
`accept` returns at once. No concurrency, so neither depends on #122.

They assert bytes that crossed, not numbers that came back:

| assertion | why a simulation cannot satisfy it |
|---|---|
| received payload == sent payload, **both directions** | a `send` that ignores its buffer transmits nothing |
| a second `bind` to the same address **fails** | a simulated bind cannot make a port unavailable |
| `accept` returns a **different** fd, and it carries the client's bytes | a number ≥ 0 proves nothing about which socket it is |
| after `shutdown`, the peer reads **EOF** | a no-op shutdown leaves the peer waiting |

### Verified by mutation

Reverting `Sys·socket` to the counter and re-running:

```
P1_041 on the MUTANT:
  socket: OK
  connect: OK
  send: OK                          ← the count assertion still passes
  payload arrived intact: FAILED    ← the payload assertion does not
  reply arrived intact: FAILED

P1_042 on the MUTANT:
  bind: OK
  listen: OK
  port really taken: FAILED         ← the side-effect assertion catches it
```

**`send: OK` passing on the mutant is the point.** The simulation returns the
plausible number, so a test asserting `sent == 9` inherits the defect it was
meant to catch — which is exactly what the old file did. Only the payload and
side-effect assertions discriminate.

## Not a regression: `P0_003_websocket_real`

It timed out in one suite run and I nearly reported it as breakage from this
change. Sampled 3× on each build, it passes 3/3 both with and without — the
timeout was transient contention. One red run is no more evidence than one
green one.

## Tests

Measured on `develop` (652dc58) first, same machine, **default features**,
`--no-fail-fast`:

| | develop @ 652dc58 | this branch |
|---|---|---|
| `cargo test --release --no-fail-fast` (lib) | 590 passed / 2 failed | 590 passed / 2 failed |
| `jormungandr/tests/run_tests_rust.sh` | 797 / 4 failed / 6 skipped | 797 / 4 failed / 6 skipped |

`llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}`; the 4
suite failures are the Kafka/AMQP broker tests.

Identical totals — `P1_041` and `P1_042` were passing before and pass now, but
they were passing against a simulation and now pass against a socket.

*(Feature set stated deliberately: `--no-default-features --features
minimal,protocols` omits all 47 `llvm_codegen` tests including both known reds,
so it reports a clean green that is not comparable to the above.)*

### DIFF
--- a/jormungandr/tests/spec/22_native_runtime/P1_041_socket_connect.expected
+++ b/jormungandr/tests/spec/22_native_runtime/P1_041_socket_connect.expected
@@ -1,6 +1,8 @@
 socket: OK
 connect: OK
 send: OK
-recv: OK
+payload arrived intact: OK
+reply arrived intact: OK
 shutdown: OK
+peer saw EOF: OK
 close: OK
--- a/jormungandr/tests/spec/22_native_runtime/P1_041_socket_connect.sg
+++ b/jormungandr/tests/spec/22_native_runtime/P1_041_socket_connect.sg
@@ -1,54 +1,57 @@
-// Test socket connect operations
-// This tests the native runtime networking syscalls
+// Test: socket connect, send, recv, shutdown
+// Spec: 22-NATIVE-RUNTIME.md
 // Priority: P1
+//
+// A real connection over loopback, with the payload compared rather than
+// counted. This file used to say "in interpreter mode, this always succeeds"
+// and "Send some data (simulated)" in its own comments: `Sys·send` returned
+// the length it was handed without reading the buffer, and `Sys·recv` returned
+// 0, so `sent == 5` and `received >= 0` passed while nothing crossed a socket
+// (#124, #167).
+//
+// Asserting a plausible number is exactly the defect being fixed, so every
+// assertion here is about bytes that arrived.
 
 rite main() {
-    // Test: Create and connect a TCP socket (simulated)
-    ≔ sock = Sys·socket(AF_INET(), SOCK_STREAM(), 0);
+    ≔ srv = Sys·socket(2, 1, 0);
+    Sys·setsockopt(srv, 1, 2, 1, 4);
+    Sys·bind(srv, "127.0.0.1:7432", 0);
+    Sys·listen(srv, 8);
 
-    ⎇ sock >= 0 {
-        println("socket: OK");
+    ≔ cli = Sys·socket(2, 1, 0);
+    ⎇ cli < 0 { println("socket: FAILED"); ⤺; }
+    println("socket: OK");
 
-        // Connect (in interpreter mode, this always succeeds)
-        ≔ result = Sys·connect(sock, 0, 16);
-        ⎇ result == 0 {
-            println("connect: OK");
+    ⎇ Sys·connect(cli, "127.0.0.1:7432", 0) == 0 {
+        println("connect: OK");
+    } ⎉ { println("connect: FAILED"); ⤺; }
 
-            // Send some data (simulated)
-            ≔ sent = Sys·send(sock, 0, 5, 0);
-            ⎇ sent == 5 {
-                println("send: OK");
-            } ⎉ {
-                println("send: FAILED");
-            }
+    ≔ peer = Sys·accept(srv, 0, 0);
 
-            // Receive (returns 0 ∈ interpreter - EOF)
-            ≔ received = Sys·recv(sock, 0, 1024, 0);
-            ⎇ received >= 0 {
-                println("recv: OK");
-            } ⎉ {
-                println("recv: FAILED");
-            }
+    // Client to server. The count must be the real count *and* the bytes must
+    // be the bytes: a send that reports its argument without transmitting
+    // passes the first check and fails the second.
+    ≔ sent = Sys·send(cli, "PING-7432", 9, 0);
+    ⎇ sent == 9 { println("send: OK"); } ⎉ { println("send: FAILED"); }
+    ⎇ Sys·recv_string(peer, 64) == "PING-7432" {
+        println("payload arrived intact: OK");
+    } ⎉ { println("payload arrived intact: FAILED"); }
 
-            // Shutdown
-            ≔ shut = Sys·shutdown(sock, SHUT_RDWR());
-            ⎇ shut == 0 {
-                println("shutdown: OK");
-            } ⎉ {
-                println("shutdown: FAILED");
-            }
-        } ⎉ {
-            println("connect: FAILED");
-        }
+    // Server to client, so the direction is not the thing that happens to work.
+    Sys·send(peer, "PONG", 4, 0);
+    ⎇ Sys·recv_string(cli, 64) == "PONG" {
+        println("reply arrived intact: OK");
+    } ⎉ { println("reply arrived intact: FAILED"); }
 
-        // Close
-        ≔ closed = Sys·close(sock);
-        ⎇ closed == 0 {
-            println("close: OK");
-        } ⎉ {
-            println("close: FAILED");
-        }
-    } ⎉ {
-        println("socket: FAILED");
-    }
+    // A shutdown the peer can observe. After the client stops writing, the
+    // server reads EOF -- empty, not blocked, not the previous payload again.
+    // A no-op shutdown leaves the peer waiting instead.
+    ⎇ Sys·shutdown(cli, 1) == 0 { println("shutdown: OK"); } ⎉ { println("shutdown: FAILED"); }
+    ⎇ Sys·recv_string(peer, 64) == "" {
+        println("peer saw EOF: OK");
+    } ⎉ { println("peer saw EOF: FAILED"); }
+
+    Sys·close(peer);
+    Sys·close(srv);
+    ⎇ Sys·close(cli) == 0 { println("close: OK"); } ⎉ { println("close: FAILED"); }
 }
--- a/jormungandr/tests/spec/22_native_runtime/P1_042_socket_server.expected
+++ b/jormungandr/tests/spec/22_native_runtime/P1_042_socket_server.expected
@@ -1,6 +1,9 @@
 socket: OK
-setsockopt: OK
 bind: OK
 listen: OK
+port really taken: OK
+connect: OK
 accept: OK
+accept gave a new fd: OK
+accepted socket carries the data: OK
 close: OK
--- a/jormungandr/tests/spec/22_native_runtime/P1_042_socket_server.sg
+++ b/jormungandr/tests/spec/22_native_runtime/P1_042_socket_server.sg
@@ -1,57 +1,61 @@
-// Test socket server operations
-// This tests bind, listen, accept syscalls
+// Test: socket server operations — bind, listen, accept
+// Spec: 22-NATIVE-RUNTIME.md
 // Priority: P1
+//
+// A real server socket on loopback. This file used to pass `0` as the address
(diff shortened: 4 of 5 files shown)


paraphilic-ecchymosis:
Merged at `5be7556` — head confirmed identical to the verified commit immediately before merging.

Independently verified by lr-adhoc-16, and the method is why the result means anything.

## The instrument had to be external, and was

The defect being fixed is `Sys·send` returning the length of bytes it **never read and never transmitted**. So any evidence drawn from the API's own return values would have inherited the very defect it was checking for — asserting `Sys·send` returned 9 and `Sys·recv` returned 9 is asking the defect whether it is still present.

What was measured instead:

- **`ss` reporting a `LISTEN` owned by the sigil pid at fd 3** — the operating system's account, not the program's.
- **A separate Python process actually receiving the bytes** — a foreign peer, which a simulation cannot satisfy.

And the negative control that makes those facts evidence rather than decoration: **the pre-fix build failed both while reporting success on every call.**

The three socket specs (`P1_040_socket_basic`, `P1_041_socket_connect`, `P1_042_socket_server`) were run against the **merge result** — parents `5be7556 + d226926` — not the branch head, and each matched its `.expected` output.

## A correction the verifier made to its own report

An earlier note said this PR was "+4 tests over the 635/637 baseline". That was wrong, and wrong in the exact way this repo has spent the day cataloguing: the 635 came from `2a8f1a9`, while this PR's merge base is `652dc58`, several commits later. **Those four tests came from `develop` moving, not from this PR.**

`#169` adds **zero** Rust unit tests — correctly, because its tests are jormungandr `.sg` spec files, which `cargo test` does not run. That is also why the specs had to be executed separately. Against `develop` tip `d226926`: 641 passed / 2 failed / 1 ignored, and the merge result is identical. The two failures are the standing `llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}`.

Merging this changes the cargo-test count by zero.

## Not measured

- The jormungandr spec suite as a whole — three socket specs were run on the merged tree, not the rest of `tests/spec/`.
- The behaviour of the four gap commits (`--version` provenance, `#163`, `#166`) — they were present in what was built, but none of their subject matter was exercised.
- `epoll_*` (scoped to #122 by this PR), the non-native/WASM simulation path, and `Sys·recv` on binary payloads.

Issue #82, opened 2026-09-10
S12: `sigil migrate`'s keyword-identifier pass renames on a false premise


paraphilic-ecchymosis:
`sigil migrate`'s keyword-identifier pass (19a0f8c) renames every keyword it finds in a binding position, on the stated premise that **"a keyword can never legitimately be an identifier."**

That premise is false. The parser accepts most keywords as struct field names, in both the declaration and the literal.

## Reproduction

Verified against a release binary built from `develop`:

```sigil
Σ Accepted {
    anima: Int,
    state: Int,
    body: Int,
    layer: Int,
    scope: Int,
}

rite main() {
    ≔ a = Accepted { anima: 1, state: 2, body: 3, layer: 4, scope: 5 };
}
```

```
$ sigil check s12-accept.sigil
✓ s12-accept.sigil - no errors
```

All five are keyword tokens in the lexer, and all five parse fine in field position. The pass renames every one of them anyway.

Only two keywords are genuinely rejected there:

```
Σ Rejected { aspect: Int }   →  [E0002] Unexpected token: expected identifier, found Trait
Σ Rejected { alter: Int }    →  [E0002] Unexpected token: expected identifier, found Alter
```

So the pass is correct for `aspect` and `alter`, and wrong for the rest.

## Observed damage

Running migrate over this repository rewrote **nine existing entries** in `CONCLAVE.sigil`, changing `anima` to `anima_` — in a file whose own header reads:

> APPEND-ONLY: Never modify existing entries.

That rewrite is the *entire* difference between the two copies of `CONCLAVE.sigil` that appeared during #62's merge. Both sides had appended the identical 49 lines; nothing else differed. The conflict existed solely because of this pass.

A rename that buys no parseability is a workaround, not a fix — and it silently violated a stated invariant of the file it edited.

## Expected behaviour

The pass should consult which keywords the parser actually rejects **in the position the name was found**, rather than renaming all of them. Position matters: a keyword rejected in one position may be accepted in another, which is exactly what this bug demonstrates.

## Probe file

A minimal case is committed on `claude/lares-spec-sigil-rebuild-2cqmp3`:

```
docs/findings/lares-probe-2026-09-07/s6-s11/S12_keyword_field_names.sigil
```

## Provenance

[...]

PR #171, opened 2026-09-10
fix(parser): let `of` and `each` be names, like the other 90 keywords

### LINKED ISSUE #143: `of` and `each` are rejected as field names; the other 90 keywords in `keyword_as_ident` are accepted
`of` and `each` are the only two keyword tokens that cannot be used as a field name. All 90 in `keyword_as_ident` can.

## Reproduction

Release binary from `develop` @ `aa3d039`, `--features minimal,protocols,react-migrate,wasm`. One line each, real exit codes:

```sigil
☉ rite f(p: Any) -> Any! { ⤺ p.of; }     // exit 1  ← parse error
☉ rite f(p: Any) -> Any! { ⤺ p.each; }   // exit 1  ← parse error
☉ rite f(p: Any) -> Any! { ⤺ p.in; }     // exit 0
☉ rite f(p: Any) -> Any! { ⤺ p.for; }    // exit 0
☉ rite f(p: Any) -> Any! { ⤺ p.type; }   // exit 0
☉ rite f(p: Any) -> Any! { ⤺ p.ref; }    // exit 0
```

An object literal key is affected too: `{ of: 1 }` is a parse error.

## Cause

`lexer.rs` gives both an ASCII prose alias onto a symbol token:

```rust
#[token("∀")]
#[token("each")]   // Prose alternative for impl blocks: ⊢ Trait each Type
ForAll,

#[token("∈")]
#[token("of")]     // Prose alternative: each x of iter { }
ElementOf,
```

`parser.rs::keyword_as_ident` (line 9684) maps 90 keyword tokens back to their spelling so they can appear as field/variable names. `ForAll` and `ElementOf` are not among them — the only two omissions.

## Why this is the parser's bug and not the migrator's

#82 closed on the finding that *"a keyword can never legitimately be an identifier"* is a false premise, and stopped `migrate` renaming keywords in field position, because the parser accepts them. That premise is right for 90 tokens and wrong for these two, which is exactly the shape that makes it hard to notice.

`Token::ElementOf => Some("of")` and `Token::ForAll => Some("each")` make the set uniform. (The symbol spellings `∈` / `∀` never reach field position, so there is no round-trip ambiguity in practice.)

## How it was hit

React source in `paraphilic-ecchymosis/lares` has a component prop literally named `of`:

```tsx
function FacetRail(props: {
  counts: TodoItem[]
  of: (it: TodoItem) => string | null | undefined
  …
})
```

`sigil migrate` renamed the destructured binding to `of_` but left the property access, emitting:

```sigil
≔ of_ = self.props.of;
```

which does not parse. That is a second, smaller defect — the rename is applied at the binding and not at the access — but it only matters because the parser rejects `.of`. With `keyword_as_ident` fixed, the emitted code is correct as written.

**Blast radius:** one component fails to parse, and because a project build is one module, `sigil wasm project/` fails for the whole 106-component client on this single field access.

## Diagnostic note

The error names the token, not the source text:

```
[E0002] Error: Unexpected token: expected identifier, found ElementOf at 35..37
```

The file contains no `∈`. I grepped for one, found none, and concluded the error had been attributed to the wrong file before spotting that `of` *is* `ElementOf`. Naming the lexeme — `found keyword `of`` — would have made this immediate. Same reporting theme as #131.

paraphilic-ecchymosis:
`keyword_as_ident` maps keyword tokens back to their spelling so they can be field, parameter or local names. It covered **90 of the 92**. The two it missed were `ElementOf` and `ForAll` — and the lexer gives both an ASCII prose alias:

```rust
#[token("")] #[token("of")]    ElementOf
#[token("")] #[token("each")]  ForAll
```

So `of` and `each` were the only two keywords that could not be a name:

```sigil
 rite f(p: Any) -> Any! {  p.in; }    // exit 0
 rite f(p: Any) -> Any! {  p.for; }   // exit 0
 rite f(p: Any) -> Any! {  p.type; }  // exit 0
 rite f(p: Any) -> Any! {  p.ref; }   // exit 0
 rite f(p: Any) -> Any! {  p.of; }    // exit 1  <- parse error
```

## Why it survived

This is the residue of **#82**, which closed on the finding that *"a keyword can never legitimately be an identifier"* is a false premise, and stopped `migrate` renaming keywords in field position. That premise is right for 90 tokens and wrong for two — exactly the shape that survives review.

## How it was found

The Lares/Qliphoth capability probe. React's `FacetRail` has a prop literally named `of`, so `migrate` emitted `= of_ = self.props.of;` — the binding renamed, the access not — and **one field access failed the whole 106-component project link**, since a project build is one module. With this fix that file parses and the link moves on to the next blocker.

Both keyword spellings keep working: `∀ x of xs` and `⊢ Trait each Type` still parse. The symbol forms never reach name position, so the mapping back is unambiguous.

## Two corpus tests updated, both asserting the old behaviour

- **`a_closure_parameter_is_a_binding`** expected `of` reported as a collision, on the stated premise that it *"is refused in every position, so it is a true positive"*. No longer true: `of` now binds cleanly in field, parameter and local position, so it joins `this` in not being reported.

  Worth noting the detector needed **no change**. It probes the real parser once per `NamePosition`, so it tracked the behaviour change on its own — only the hard-coded expectation was stale. That is the design working.

- **`a_comment_does_not_lose_the_enclosing_brace`** used `each` as its specimen. Its subject is the bracket stack surviving a comment, not `each`, so the specimen is now `loop`, which is still refused in field position. Otherwise the test would be asserting the absence of a defect that was fixed rather than the behaviour it exists for.

## Tests

Provenance: **`sigil-lang` `develop` @ `652dc58`, DEFAULT features, `cargo test --release --no-fail-fast`.**

| | all binaries |
|---|---|
| this branch | **640 passed, 2 failed** |
| baseline `652dc58` | 639 passed, 2 failed |

The +1 is the regression test. The two failures are `llvm_codegen::llvm::tests::test_generic_struct_basic` and `test_generic_struct_two_params` — **pre-existing, identical before and after**.

Default features used deliberately: a minimal-flags build does not compile `llvm_codegen` at all, which is exactly where both failures live, so its clean 0 is a false green. And LLVM 18.1.3 is present on this machine, so there is no reason to reach for minimal flags in the first place.

A regression test pins both halves — the five name positions that were rejected, and the two keyword uses that must keep parsing.

Closes #143

 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/parser/src/main.rs
+++ b/parser/src/main.rs
@@ -7988,9 +7988,14 @@ mod migrate_span_tests {
         // its bracket stack per span sees no enclosing brace here, takes no
         // position, and reports nothing — so it misses the documented field,
         // which is the ordinary shape of the thing this diagnostic is for.
+        // Specimen is `loop`, which is still refused in field position. It used
+        // to be `each` — an ASCII prose alias for `∀` that the parser now
+        // accepts as a field name, so it stopped being a collision and this
+        // test would otherwise be asserting the absence of a defect that was
+        // fixed rather than the bracket-stack behaviour it is for.
         assert_eq!(
-            f("pub struct S {\n    /// Emit extend method.\n    pub each: Option<Each>,\n}"),
-            vec![("each".into(), 1)]
+            f("pub struct S {\n    /// Emit extend method.\n    pub loop: Option<Loop>,\n}"),
+            vec![("loop".into(), 1)]
         );
         assert_eq!(
             f("struct S {\n    pub n: u8, // trailing\n    pub aspect: f32,\n}"),
@@ -8678,14 +8683,22 @@ mod collision_corpus_tests {
     fn a_closure_parameter_is_a_binding() {
         // #84/#89. `this` binds cleanly, so no closure parameter spelled
         // `this` may be reported -- develop carried exactly that as a live
-        // false positive, in cranelift-isle and rustc-demangle. `of` is
-        // refused in every position, so it is a true positive and stays.
+        // false positive, in cranelift-isle and rustc-demangle.
+        //
+        // `of` was here too, on the premise that it "is refused in every
+        // position, so it is a true positive". That premise is no longer true:
+        // `of` is the ASCII prose alias for `∈`, and it now resolves as an
+        // ordinary name in field, parameter and local position like the other
+        // 90 keywords. So it binds cleanly and joins `this` in not being
+        // reported. The detector needed no change to learn this — it probes the
+        // real parser per position, which is exactly why it tracked the change
+        // on its own and this expectation did not.
         //
         // `tome` is counted twice: once for the typed `|mut tome: u32|`,
         // found by its colon, and once for the untyped `|mut tome|` beside it.
         // The untyped form was #91 -- the byte before the name is the `t` of
         // `mut`, so the test had to look one word further back for the bar.
-        expect("closure_params.rs", &[("of", 1), ("tome", 2)]);
+        expect("closure_params.rs", &[("tome", 2)]);
     }
 
     #[test]
--- a/parser/src/parser.rs
+++ b/parser/src/parser.rs
@@ -9684,6 +9684,22 @@ impl<'a> Parser<'a> {
     fn keyword_as_ident(token: &Token) -> Option<&'static str> {
         match token {
             // Common keywords that may be used as field/variable names
+            //
+            // `of` and `each` are the ASCII prose aliases the lexer gives to
+            // `` (ElementOf) and `` (ForAll). They were the only two of the 92
+            // keyword tokens missing from this table, which made them the only
+            // two that could not be a field name: `p.in`, `p.for`, `p.type` and
+            // `p.ref` all parse, and `p.of` does not. #82 closed on the finding
+            // that "a keyword can never legitimately be an identifier" is a
+            // false premise and stopped `migrate` renaming keywords in field
+            // position — true for 90 tokens and false for these two, which is
+            // what let it survive. React's `FacetRail` has a prop named `of`,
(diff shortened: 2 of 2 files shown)


paraphilic-ecchymosis:
Merged at `ace2f85`.

**Not merged on CI.** This repo's `ci.yml` carries `continue-on-error: true` on four steps including Run tests, and the raw logs show `Process completed with exit code 101` swallowed on both runners — see #175. The green tick here answers "did the job step complete", not "did the tests pass", so it is worth nothing either way on this PR.

What it merged on instead — independent local verification, by execution:

**Unit level.** No regression, no new failures, and the new `parser::tests::of_and_each_are_usable_as_names` passes.

**Integration level, and this is the part that makes it convincing.** One session predicted a falsifiable consequence; a different session tested it without having made the prediction. Binaries built with `react-migrate,wasm` at this PR's head **and at its parent**, then the real lares-client `project/` regenerated against `lares@634034d` read-only:

```
BEFORE  facet_rail.sigil → exit 1
        expected identifier, found ElementOf at 903..905
AFTER   exit 0
        full link advances to exactly the predicted next gap:
        2 name(s) resolved to nothing: filter, now      (already filed, #145)
```

No hand patch. A 106-module project link is a far stronger test of a keyword change than a unit test, because it exercises the keyword in every position a real corpus puts it in — and the prediction naming *which* gap would surface next is what makes the result falsifiable rather than merely positive.

The prediction came from lr-adhoc-17 while verifying lares-client #3; the measurement from lr-adhoc-20, told the prediction but not how to satisfy it.

**Not measured:** clippy and fmt (both also `continue-on-error` upstream, not re-run locally), and the macOS build.

Issue #175, opened 2026-09-10
CI is green regardless: four steps carry continue-on-error, and the test step exits 101 while the job passes


paraphilic-ecchymosis:
Measured 2026-09-10 by an independent verifier, confirmed directly against `origin/develop`.

## The workflow does not gate on anything

`.github/workflows/ci.yml`, "Build and Test":

```yaml
- name: Run tests
  run: cargo test --release --no-default-features --features jit,llvm,wasm,lsp,protocols
  continue-on-error: true          # line 65

- name: Check formatting
  run: cargo fmt -- --check
  continue-on-error: true          # line 70

- name: Run clippy
  run: cargo clippy … -- -D warnings
  continue-on-error: true          # line 75

- name: sigil check examples
  continue-on-error: true          # line 83
```

**Four steps, all non-gating.** The raw log shows `Process completed with exit code 101` — cargo test's real-failure code — on **both** the ubuntu and macOS runs, swallowed. The job reports success.

> The green checkmark answers **"did the job step complete"**, not **"did the tests pass."**

There is also **no `--no-fail-fast`**, so the run stops at the first failing binary and `main.rs`'s tests never execute at all. `lib` alone reports 822 passed / 2 failed.

## Why this matters beyond tidiness

Daemon's standing rule is that free CI must be running and green before merging, and sigil-lang is public so its Actions are free. That rule is currently satisfiable by a workflow that cannot fail.

**I merged five PRs today citing this green** — #154, #160, #163, #168, #169. Each also had independent local verification by a separate session, and those verifications are what the merges actually rest on. But I wrote "CI green on all four jobs" in merge comments as though it were evidence, and it was not. Those comments are being corrected.

## A third incomparable flag set

CI builds with `--no-default-features --features jit,llvm,wasm,lsp,protocols`. That is neither `default` nor the `minimal,protocols[,react-migrate,wasm]` sets already in circulation, so **CI's test count is not comparable to any figure quoted in a PR body or an issue here.** See the flags note: `minimal,protocols` is a strict subset of default missing the 47 `llvm_codegen` tests where both known failures live, and the 4-flag set is mutually non-superset with default.

## Fix

1. **Remove `continue-on-error` from the test step.** If the two standing `llvm_codegen::llvm::tests::test_generic_struct_{basic,two_params}` failures are why it is there, that is an argument for an explicit expected-failure list or a filter, not for a gate that cannot fail. A known red should be *named*, not blanket-tolerated.
[...]

Issue #179, opened 2026-09-11
parse_manifest has_lib/has_bin detection ignores .sg — sigil build cannot build any .sg-only crate


paraphilic-ecchymosis:
`parse_manifest`'s has_lib/has_bin detection only recognizes `.sigil`, never `.sg` — `sigil build` cannot build any `.sg`-only crate

## What happened

Reproduced against `origin/develop` @ `d6e27270e7a5a03940c1cd7f6457d39a39ca1b46`, default
features, on a single crate with a correctly-cased manifest and a real library entry point:

```
$ ls src/
lib.sg

$ cat sigil.toml
[project]
name = "amdusias-core"
version = "0.1.0"

$ sigil build
Error: no src/lib.sigil or src/main.sigil found
A tome must have at least one of these files.
```

The crate unambiguously has a library entry point — `src/lib.sg` — and the manifest is
lowercase (isolated from #176 deliberately, to attribute this failure to a distinct cause).
`sigil build` reports it as having neither a lib nor a bin anyway.

## Root cause

`parse_manifest` (parser/src/main.rs:4622-4623) hardcodes the `.sigil` extension with no
fallback:

```rust
let has_lib = manifest_dir.join("src/lib.sigil").exists();
let has_bin = manifest_dir.join("src/main.sigil").exists();
```

This is inconsistent with the rest of the same file, which explicitly treats `.sigil` and
`.sg` as equally valid source extensions in several other places — workspace file discovery
(~line 981-982, 1147, 2009-2010, 2089) and module sort ordering (~661-671, 913-921,
1015-1038) all check both. `build_project`'s manifest reader appears to predate `.sg` being
added as a recognized extension elsewhere, and was never updated.

## Impact

Blocks `sigil build` on every `.sg`-only project, independent of #176 and #177 — even once
both of those are fixed, a project with only `.sg` source files (never `.sigil`) still fails
here. amdusias is 100% `.sg` (51 files, 0 `.sigil`), so it hits this unconditionally.

This is the second of the two skip-causes named in #177: once #176 is fixed, every member of
a `.sg`-based workspace will fail here instead — and, per #177, that failure currently gets
reported as a silent overall success.

## Suggested fix

Have `parse_manifest`'s has_lib/has_bin check accept both extensions, matching the pattern
already used elsewhere in this file:

```rust
let has_lib = manifest_dir.join("src/lib.sigil").exists() || manifest_dir.join("src/lib.sg").exists();
[...]

Issue #178, opened 2026-09-11
sigil check: never accepts a directory argument, in any published version or on develop — ci.yml's "sigil check ." cannot pass under any pin


paraphilic-ecchymosis:
## What happened

`sigil check` has never accepted a directory argument. It takes exactly one
positional argument, read as a single file via `fs::read_to_string(path)`, with
no `is_dir()` branch anywhere in `check_file`. `sigil check .` — the natural
invocation for "check my whole project" and what `Daemoniorum-LLC/amdusias`'s
`ci.yml` actually runs — fails on **every** version, published or not.

## Reproduction across the full published + unreleased range

```
$ git show v0.3.0:parser/src/main.rs | sed -n '230,252p'      # oldest active published version
        "check" => {
            ...
            check_file(&args[2], format, quiet, apply_fixes)
        }
$ git show v0.3.0:parser/src/main.rs | grep -n '^fn check_file' 
1601:fn check_file(path: &str, format: OutputFormat, quiet: bool, apply_fixes: bool) -> ExitCode {
$ git show v0.3.0:parser/src/main.rs | sed -n '1601,1605p'
fn check_file(path: &str, format: OutputFormat, quiet: bool, apply_fixes: bool) -> ExitCode {
    let source = match fs::read_to_string(path) {
        ...
```

```
$ git show v0.4.0-rc.4:parser/src/main.rs | sed -n '243,253p'   # newest ACTIVE published version
        "check" => {
            ...
            check_file(&args[2], format, quiet, apply_fixes)
```

```
$ git show d6e2727:parser/src/main.rs | sed -n '346,365p'   # origin/develop HEAD, 193 commits past the newest tag — the newest code anywhere
        "check" => {
            ...
            check_file(&args[2], format, quiet, apply_fixes, strict)
```

`check_file` at `d6e2727` still has no directory branch — same `fs::read_to_string(path)`, same
usage string `Usage: sigil check <file.sigil> [...]`.

For contrast, `is_dir()` handling **does** exist elsewhere in the same file — `lint`, `build`,
`build_workspace`, `test`/`run_tests_workspace`, `migrate --workspace` all walk directories. `check`
alone has never gained it, at any point in this project's git history.

## Impact

`Daemoniorum-LLC/amdusias`'s `ci.yml` runs `sigil check .` as its Type Check job. That step cannot
[...]

Issue #177, opened 2026-09-11
build_workspace exits 0 having built nothing — a fully skipped workspace reports success


paraphilic-ecchymosis:
`build_workspace` exits 0 having built nothing — a skipped workspace is reported as a successful build

## Summary

When every member of a workspace is skipped, `build_workspace` returns
`ExitCode::SUCCESS`: **0/0 crates, zero artifacts produced, exit code 0.**

Reproduced concretely against `amdusias` (7 crates, 51 `.sg` files, 17,283 lines):
a full workspace build skipped every member and reported success.

## Why this one matters more than the bugs that trigger it

An absent artifact fails loudly. This passes quietly. Any CI step, script, or
agent that checks the exit status — which is the normal thing to check — is told
the build succeeded, and the only way to notice is to go looking for a binary
that isn't there.

It also makes the two bugs below invisible instead of obvious. They are ordinary
bugs; this turns them into a silent one.

## The two skips that trigger it here

1. **Manifest case sensitivity** — the build only accepts lowercase `sigil.toml`.
   Already filed as #176.
2. **`parse_manifest`'s lib/bin detection only ever checks for `.sigil` files,
   never `.sg`.** So even a correctly-cased manifest fails on every one of
   amdusias's source files, because they are all `.sg`.

Either one alone causes every member to be skipped, and the skip is then reported
as success.

## Suggested fix

A build that produced no artifacts is not a successful build. Distinguish
"nothing to do" from "everything was skipped", and make the latter a non-zero
exit naming the crates and the reason each was skipped. If a zero-member
workspace should legitimately succeed, say so explicitly in the output rather
than letting it share an exit code with a total skip.

## Acceptance — must be able to fail

- A workspace where every member is skipped exits **non-zero** and names why.
- A workspace that genuinely builds N crates still exits 0 and produces N
  artifacts — assert the artifacts exist, not just the status.
- A **mixed** workspace (some build, some skip) does not report plain success:
  the skipped members are reported. Without this case, a fix that only handles
  all-or-nothing still hides partial skips.

Found while establishing a build baseline for amdusias (Lares LARES-343 Ph1).
The session deliberately did not rename `Sigil.toml` or `.sg` files to work
around #176 and the detection bug, so the false pass was observed rather than
papered over.

paraphilic-ecchymosis:
[...]

PR #180, opened 2026-09-11
fix(build): stop a fully-skipped workspace build from exiting 0

paraphilic-ecchymosis:
## Summary

Fixes the four build-subsystem bugs (#176, #177, #178, #179) that stop `sigil build` from working on a real repository (amdusias), which blocks LARES-343 and therefore Orpheus.

- **#177 (fixed first — see rationale below)**: `build_workspace` returned `ExitCode::SUCCESS` when every workspace member was skipped — 0/0 crates, zero artifacts, exit 0. A CI step or an agent checking only the exit status was told the build succeeded. Now:
  - a workspace where every member is skipped exits non-zero and names why
  - a workspace that fails to build any member exits non-zero and names why
  - a mixed workspace (some built, some skipped) reports the skipped members by name and reason, and exits non-zero rather than reporting plain success
- **#176**: `sigil build` only ever looked for lowercase `sigil.toml`, while `test` and project-detection already accepted `Sigil.toml`. `build_project` and `build_workspace` now both resolve the manifest via a shared `find_manifest_path` helper that checks both spellings.
- **#179**: `parse_manifest`'s `has_lib`/`has_bin` detection only ever looked for `.sigil` files, never `.sg` — so a correctly-cased manifest still failed on every amdusias source file (all 51 are `.sg`). Detection and the actual build step (previously a hardcoded `src/lib.sigil` / `src/main.sigil` path) now both go through a `resolve_source_file` helper that checks `.sigil` then `.sg`.
- **#178**: `sigil check` never accepted a directory argument. It now walks `.sigil`/`.sg` files recursively (`collect_check_files`) and checks each one, aggregating the exit code; an empty directory is a named error rather than a silent pass.

### Why #177 first

The other three are ordinary bugs that produce a visible failure. #177 turned them into silent ones: fixing #176/#178/#179 without #177 would still report exit 0 with nothing built. Fixing #177 first means the rest produce observable progress instead of more zeroes that look like success.

## Verification

- **Rust unit tests**: added 11 tests (`build_subsystem_tests` in `parser/src/main.rs`) covering `find_manifest_path`, `resolve_source_file`, `collect_check_files`, and `build_workspace`'s exit-code behavior for all-skipped / zero-declared-members / missing-manifest cases. `cargo test --release --bin sigil`: 60/60 passing (49 pre-existing + 11 new).
- **Full language test suite** (`jormungandr/tests/run_tests_rust.sh`, full default features incl. `llvm`): 797/807 passing — the same 4 Kafka/AMQP infra-dependent failures documented in `docs/specs/INTERPRETER-SPEC-ROADMAP.md`, no regressions. (807 vs. the doc's 798 reflects tests added on `develop` since that doc was last updated; not part of this change.)
- **Manual, artifact-level verification** (not just exit codes):
  - Ran against **amdusias** directly: `sigil build` in amdusias now gets past manifest discovery instead of silently no-op'ing, and fails with a real, named reason (`Missing 'name' in manifest` — amdusias's `Sigil.toml` is a Cargo-style single-manifest workspace, not the per-member-manifest layout `sigil build` expects; that mismatch is amdusias's own manifest shape, outside this ticket's four bugs, and is left named rather than worked around).
  - Synthetic workspace, all members missing: exits 1, prints "Error: no workspace members were built (0/2 — all skipped)".
  - Synthetic workspace, one member builds (`Sigil.toml` + `.sg` source) and one is missing: exits 1, real LLVM binary produced and executed (`target/membera` → `hello from memberA`), skipped member reported by name and reason.
  - Synthetic workspace, both members build: exits 0, both real binaries produced and executed.
  - Single project with uppercase `Sigil.toml` + `.sg` source only: builds and the resulting binary runs correctly.
  - `sigil check <dir>`: finds a passing `.sg` file and a broken `.sigil` file in a nested subdirectory, reports both, exits 1; single-file check still works; empty directory is a named error.

## Test plan

- [x] `cargo build --release` (default features incl. `llvm`)
- [x] `cargo test --release --bin sigil`
- [x] `cd jormungandr/tests && ./run_tests_rust.sh` (full run, no regressions vs. documented baseline)
- [x] Manual runs against amdusias and synthetic fixtures (see above)

https://claude.ai/code/session_01Ceoo22g5suvZD93k7i8nDZ

### DIFF
--- a/parser/src/main.rs
+++ b/parser/src/main.rs
@@ -345,8 +345,8 @@ fn main() -> ExitCode {
         }
         "check" => {
             if args.len() < 3 {
-                eprintln!("Error: missing file argument");
-                eprintln!("Usage: sigil check <file.sigil> [--format=json|compact] [--quiet] [--strict] [--apply-suggestions]");
+                eprintln!("Error: missing file or directory argument");
+                eprintln!("Usage: sigil check <file.sigil|directory> [--format=json|compact] [--quiet] [--strict] [--apply-suggestions]");
                 return ExitCode::from(1);
             }
             // Parse format option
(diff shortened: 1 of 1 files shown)


paraphilic-ecchymosis:
## CI note: the green checkmarks on this PR are not evidence (see #175)

Run [34629357061](https://github.com/Daemoniorum-LLC/sigil-lang/actions/runs/34629357061) reports every job as success, but the "Run tests" step in both `Build and Test` jobs actually failed:

```
test result: FAILED. 824 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.6xs
##[error]Process completed with exit code 101.
```

on both `ubuntu-latest` and `macos-latest`, for the same two tests:

```
llvm_codegen::llvm::tests::test_generic_struct_basic
llvm_codegen::llvm::tests::test_generic_struct_two_params
```

That's `continue-on-error: true` on the test step (#175) swallowing the 101 and reporting the job green regardless. **Do not read the checkmarks on this PR as evidence the tests passed.**

**These two failures are pre-existing, not introduced by this change.** Confirmed directly, not assumed: I built a separate worktree at `d6e2727` (`origin/develop`'s tip, the exact commit this branch was created from, before any of this PR's commits) and ran the identical invocation CI uses:

```
LLVM_SYS_181_PREFIX=/usr/lib/llvm-18 CC=clang-18 cargo test --release \
  --no-default-features --features jit,llvm,wasm,lsp,protocols --lib
```

Result at `d6e2727`: `test result: FAILED. 824 passed; 2 failed` — same two tests, same exit code 101. Identical to this PR's commit `9a1c827`. So this PR changes nothing about that pair; they were already red on `develop` before this branch existed.

**One more thing worth flagging under #175's "no `--no-fail-fast`" point, specific to this PR:** because `--lib` fails first and there's no `--no-fail-fast`, cargo never runs the `src/main.rs` test binary in CI at all — I confirmed the ubuntu log shows only one `Running unittests ... src/lib.rs` line and zero for the bin. That means the 11 new tests this PR adds (`build_subsystem_tests`, in `parser/src/main.rs`, since `build_workspace`/`check_file`/etc. are bin-only code) **never ran in CI**, on this PR or any other PR under the current workflow. They did run locally: `cargo test --release --bin sigil` → 60/60 passing (49 pre-existing + 11 new), reported in the PR description.

Not attempting to fix #175 here — out of scope for LARES-353 — but flagging it concretely against this PR's own run rather than letting the green ticks stand unchallenged.

PR #181, opened 2026-09-11
fix(ci): gate the job on real test results, run every test binary (#175)

paraphilic-ecchymosis:
## Summary

Fixes sigil-lang#175 — CI is green regardless. Two defects, not one:

1. **`continue-on-error: true` on the "Run tests" step** swallowed cargo test's real exit code (101 on failure), so the job reported success no matter what the tests did. Removed — it's now the only step that gates the job.
2. **No `--no-fail-fast`**, so the run stopped at the first failing test binary. In practice: the `--lib` suite's two pre-existing failures (`llvm_codegen::llvm::tests::test_generic_struct_basic`/`two_params`) stopped cargo before it ever reached the `--bin sigil` binary, so `main.rs`'s own unit tests never ran in CI at all — including the 11 tests LARES-353 (#180) just added there. Added `--no-fail-fast`.

The other three steps (`fmt`, `clippy`, example-file checks) stay advisory (`continue-on-error`), which is a legitimate choice per #175's own fix list — but it's now a *stated* one: each step is renamed `... (advisory, does not gate the job)`, and a new `Advisory-steps summary` step writes to the job summary which steps are advisory, instead of one undifferentiated green tick.

Feature-flag alignment (#175's 4th point — CI's `jit,llvm,wasm,lsp,protocols` vs. `default`/`minimal,protocols`) is intentionally **not** touched here: changing what gets tested is a separate decision from making the gate honest, and bundling it in risks hiding one change inside the other.

## ⚠️ Expect this PR's own CI run to go RED, and that is correct

Once the gate is real, this branch's **own** two pre-existing `llvm_codegen` failures (see above) will surface as a red `Build and Test` job on *this PR*. **That is the fix working, not a regression** — those two tests have been failing on `develop` all along; `continue-on-error` was the only thing hiding it. Do not revert this PR because its check is red; read the log.

## Proof (the AC that matters): a deliberately-failing test turns the job RED

Commit `3bf7afe` added a temporary, obviously-failing test (`ci_gate_probe_deliberately_fails`, `assert!(false, ...)`) specifically to observe this gate catch something in a real Actions run rather than assert it works. Once that run is confirmed red *because of the gate* (not swallowed), the probe commit will be reverted before merge — see the follow-up commit / PR update.

CI passing today (on `develop`, pre-#175-fix) proves nothing, because CI already passes today with two real, live test failures. This PR's job going red — first from the deliberate probe, then, after the probe is removed, from the two real pre-existing failures — is the actual proof.

## Test plan

- [x] `cargo test --release --bin sigil` locally: probe fails as expected (`0 passed; 1 failed`)
- [ ] This PR's Actions run on GitHub goes RED (both ubuntu and macOS) because of the probe — to be confirmed and linked here
- [ ] Probe commit reverted, PR updated to note the run went red for the two known pre-existing `llvm_codegen` failures (also correct, also not a regression)

https://claude.ai/code/session_01Ceoo22g5suvZD93k7i8nDZ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,28 +59,51 @@ jobs:
         working-directory: parser
         run: cargo build --release --no-default-features --features jit,llvm,wasm,lsp,protocols
 
+      # Gating: a red here must fail the job. No continue-on-error, and
+      # --no-fail-fast so a failure in one test binary (e.g. the `--lib`
+      # suite) doesn't prevent later binaries (e.g. `--bin sigil`, which
+      # carries its own unit tests) from running at all. See #175: before
+      # this, continue-on-error swallowed the real exit code and the absence
+      # of --no-fail-fast meant the bin crate's tests never even ran.
       - name: Run tests
         working-directory: parser
-        run: cargo test --release --no-default-features --features jit,llvm,wasm,lsp,protocols
-        continue-on-error: true
+        run: cargo test --release --no-default-features --features jit,llvm,wasm,lsp,protocols --no-fail-fast
 
-      - name: Check formatting
+      # The steps below are deliberately advisory (continue-on-error): a
+      # formatting or lint nit, or a stale example, should not block a merge
+      # the way a real test failure does. That is a choice, not an oversight
+      # -- see #175 -- so each step and the job summary say so explicitly
+      # rather than leaving every step looking equally gating.
+      - name: Check formatting (advisory, does not gate the job)
         working-directory: parser
         run: cargo fmt -- --check
         continue-on-error: true
 
-      - name: Run clippy
+      - name: Run clippy (advisory, does not gate the job)
         working-directory: parser
         run: cargo clippy --no-default-features --features jit,llvm,wasm,lsp,protocols -- -D warnings
         continue-on-error: true
 
-      - name: Test example files
+      - name: Test example files (advisory, does not gate the job)
         working-directory: parser
         run: |
           ./target/release/sigil check ../examples/simple.sigil
           ./target/release/sigil check ../examples/hello.sigil
           ./target/release/sigil check ../examples/pipes.sigil
         continue-on-error: true
+
+      - name: Advisory-steps summary
+        if: always()
+        run: |
+          echo "### Advisory steps (do not gate this job)" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "Only **Run tests** gates this job. The steps below run for" >> "$GITHUB_STEP_SUMMARY"
+          echo "visibility only -- a failure in any of them is reported here" >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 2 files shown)


paraphilic-ecchymosis:
## Verified: the gate is real, and the current red is correct (not a regression)

Two runs, same branch, to isolate the two things being proven:

**Run [34632029987](https://github.com/Daemoniorum-LLC/sigil-lang/actions/runs/34632029987) (commit `3bf7afe`, with the deliberately-failing probe):** both `Build and Test` jobs actually reported **FAILURE** (not swallowed) —

```
test ci_gate_probe_deliberately_fails ... FAILED
##[error]Process completed with exit code 101.
```

on both `ubuntu-latest` and `macos-latest`. This is the AC that matters: a deliberately-failing test turns the job red. (It also surfaced a genuine bug in one of LARES-353's own new tests — see below.)

**Run [34633551223](https://github.com/Daemoniorum-LLC/sigil-lang/actions/runs/34633551223) (commit `21dfe73`, probe removed):** both jobs are **still red**, now for the real, pre-existing reason:

```
test llvm_codegen::llvm::tests::test_generic_struct_basic ... FAILED
test llvm_codegen::llvm::tests::test_generic_struct_two_params ... FAILED
test result: FAILED. 824 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
##[error]Process completed with exit code 101.
```

identically on both runners. **This red is expected and correct, not a regression** — these two tests have been failing on `develop` since before this PR existed (see LARES-353/#180's baseline measurement at `d6e2727`); `continue-on-error` was the only thing that ever hid it. Please don't revert this PR because its check is red — that red is the fix working.

`--no-fail-fast` also did what it was for: the `src/main.rs` (`--bin sigil`) test binary now runs in the same job instead of never being reached —

```
test result: ok. 60 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

60/60, on both platforms — the 49 pre-existing bin tests plus the 11 LARES-353 (#180) added, none of which had ever executed in CI before this PR.

### A real bug this exposed and fixed along the way

Running the bin tests on `macos-latest` for the first time (run 34632029987) failed one of LARES-353's own tests: `build_subsystem_tests::find_manifest_path_falls_back_to_uppercase`. Not a regression in `find_manifest_path` itself — the test assumed a case-sensitive filesystem. macOS's default APFS is case-insensitive, so `sigil.toml` and `Sigil.toml` name the same inode there; `find_manifest_path` legitimately reports back whichever spelling it checks first even when only the other was written. The manifest is still found and still reads correctly either way, which is the property that actually matters, so commit `21dfe73` rewrote the test to assert on content instead of the returned path's casing.

## Ready for review

Marking this ready — both defects in #175 are fixed and independently verified against real Actions runs at each step (not asserted from the checkmark):

1. `continue-on-error` removed from the "Run tests" step — it now gates the job for real.
2. `--no-fail-fast` added — every test binary runs regardless of an earlier one failing.
3. The three remaining advisory steps (fmt, clippy, examples) are explicitly labeled `(advisory, does not gate the job)` and summarized in the job summary, so the distinction is stated rather than implicit.

